PR #127, opened 2015-02-19
Statistical power

JWDebelius:
This is a cleaned up power notebook, loosely related to what was presented in #121 

The rendered notebook can be found [here](http://nbviewer.ipython.org/github/JWDebelius/American-Gut/blob/statistical_power/ipynb/Power.ipynb).

The text is in [Google docs](https://docs.google.com/document/d/1DOEFP4SwqiJuLrMqgGitQwhss4ayREVQ4WWueltHiRM/edit?usp=sharing), which may be a better place for text editing.

This notebook relies on code generated in #125 and on the data files generated when #126 is run.


wasade:
@JWDebelius, there are merge conflicts

wasade:
I can't link directly into the notebook, so below are the initial pass comments on the notebook itself. Note, it does not appear that the `power_plots` code or the `diversity_analysis` code is included in this PR, and which appear to be necessary for this code to work.

https://github.com/JWDebelius/American-Gut/blob/statistical_power/ipynb/Power.ipynb#L16
	- copyright 2015
	
https://github.com/JWDebelius/American-Gut/blob/statistical_power/ipynb/Power.ipynb#L27
	- css file doesn't appear to be in the repo
	
https://github.com/JWDebelius/American-Gut/blob/statistical_power/ipynb/Power.ipynb#L113
	- The return to the table of contents link is confusing as it advances you in the document. Shouldn't the TOC be early or first thing in the notebook if there is one?

function imports, american_gut.power_plots does not appear to exist nor does americangut.diversity_analysis. 

general: it is QIIME

alpha diversity parameters section, the beta diversity link is 404

In parameters for fecal samples, the ANTIBIOTIC_SELECT values are "In the past month" and "Not in the past year". Earlier in the document it indicates that "in the past week" will also included?

In the files and directories section, the download links to dropbox. Are these data not in the repo? Ah, or these precomputed distance matrices? 

In data download, we should not source from what look like random dropbox accounts. If the volume is such that it doesn't make sense to host on github, we can drop the data on ftp.microbio.me. 

In power calculation for fecal samples, the first two lines are not preallocating a list but creating an empty list. 

In power calculation for fecal samples,

	# Removes the category from the control cats, if necessary
    if cat in fecal_control_cats:        
        ctrl_cats = copy.deepcopy(fecal_control_cats)
        ctrl_cats.remove(cat)

	can be replaced with
	
	# Removes the category from the control cats, if necessary
    if cat in fecal_control_cats:        
        ctrl_cats = fecal_control_cats[:]
        ctrl_cats.remove(cat)

	but since order does not appear to matter, it would be more efficient if fecal_control_cats was a set. This is likely in a block of code that is not a bottleneck, so just a minor comment.
	
effect size estimation, the capitalization of a lot of items in the table is wrong (e.g., Ibd)

Beta diversity section has a "[beta diversity discussion]", is that a place holder?

Unweighted unifrac is quite powerful? The text previously mentioned weighted unifrac as well, should there be a plot for that too?

JWDebelius:
In the past week is a subset of in the past month. I will fix the text, but they're merged in the metadata section. But, I will update the text to reflect that.

The tables are precomputed distance matrices. I've been trying to find a better place to host them than my dropbox. My understanding was that I was to wait for Qiita to be released so the processed data could be hosted there. I'd be very happy to transfer them to some better location.

Re: fecal cats as a set: right now, the power code accepts a list because it goes into a pandas group by-object. ...Not sure if its more expensive to cast to a list every time than to copy and remove the item.

I've tried to address the capitalization by switching to labels. Im still waiting for the power notebook to run (I forgot my precomputed effect size matrices are also on the computer being repaired.)

Sorry, I will remove that placeholder.

Given the time constraints and the fact that most down stream analyses I've done have focused on unweighted UniFrac, I'd rather remove the reference to weighted UniFrac beyond a vague mention that it exists and you can use it, if you want to spend the compute. I can cite the Turnabugh paper suggesting that the presence/absence of pathway branches matters more than the core pathways and the Gordon paper arguing that rare organisms make a difference in human diseases, if that better helps justify.

wasade:
Thanks @JWDebelius, will go back over again shortly.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+    # Sets the column as the index
+    df.index = new_samples
+    del df[index_col]
+
+    # Returns the dataframe
+    return df
+
+
+def boxplot(vecs, ax=None, notch=True, interval=0.5, boxplot_props={},
```
wasade: use `None` and test instead of `{}` for defaults. The reason is that `{}` is eval'd once when the function is defined, so if it gets modified, the modification remains. For instance:

```python
>>> def foo(a, b={}):
...   b[a] = a
...   print b
...
>>> foo(10)
{10: 10}
>>> foo(100)
{10: 10, 100: 100}
>>> foo(1000)
{1000: 1000, 10: 10, 100: 100}
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        The axis where data should be plotted. If none, a new axis instance
+        will be created.
+    interval : float, optional
+        The spacing between the boxplot instances on the axes
+    notch : bool, optional
+        Displays the parametric 95% confidence interval around the mean.
+    show_counts : bool, optional
+        Shows the size of the groups below each plot on the x-axis
+    p_value : float, optional
+        Default is None. When supplied, the signfigance value will be displayed
```
wasade: significance

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    show_ygrid: bool, optional
+        Default is True. Adds horizonal lines at each major y-tick.
+    title: str, optional
+        The title to be placed on the graph.
+    ylims : list
+        The limits for the y-axis.
+    ylabel : str
+        The label text for the y-axis. Every time you leave off appropriate
+        labels and units, a science grad student grading lab reports cries
+        another bitter tear into their bottle of craft beer.
```
wasade: must be a first year as they haven't moved onto whiskey...

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    yticks : array_like
+        The positions where ticks should appear on the y-axis.
+    yticklabels : list
+        The text to be displayed at each y tick.
+    ylabel_size : int
+        The font size of the y-axis label.
+    ytick_size : int
+        The font size for the ytick labels
+    """
+
```
wasade: extra space

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    # Sets up an axes instance if necessary
+    if ax is None:
+        ax = plt.axes()
+
+    # Determines the plotting locations
+    num_cats = len(vecs)
+    xlim = [-interval/2,
```
wasade: i dont' believe this needs to be split onto two lines

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+            interval*(num_cats-1)+interval/2]
+
+    # Sets up the plotting constants
+    ticks = np.arange(0, interval*num_cats, interval)
+    counts = []
+
+    # Loops through the data
+    for idx, vec in enumerate(vecs):
```
wasade: should be able to do:

```python
for tick, vec in zip(ticks, vecs):
    ...
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        kwargs['xticks'] = ticks
+
+    _format_axis(ax, **kwargs)
+
+    return ax
+
+
+def pretty_pandas_boxplot(meta, group, cat, order=None, ax=None,
+    **boxplot_props):
+    """Creates a more attractive poxplot than pandas
```
wasade: is it a small poxplot by chance?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        `group`: A, B and C, and you only wish to compare A and C, you can
+        list order as ['A', 'C'] to limit the categories.
+    interval : float, optional
+        The spacing between the boxplot instances on the axes
+    notch : bool, optional
+        Displays the parametric 95% confidence interval around the mean.
+    show_n : bool, optional
+        Shows the size of the groups below each plot on the x-axis
+    p_value : float, optional
+        Default is None. When supplied, the signfigance value will be displayed
```
wasade: significance

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    show_xgrid: bool, optional
+        Default is False. Adds vertical lines at each major x-tick.
+    show_ygrid: bool, optional
+        Default is True. Adds horizonal lines at each major y-tick.
+    title: str, optional
+        The title to be placed on the graph.
+    ylims : list
+        The limits for the y-axis.
+    ylabel : str
+        The label text for the y-axis. Every time you leave off appropriate
```
wasade: be careful copying and pasting test

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        labels and units, a science grad student grading lab reports cries
+        another bitter tear into their bottle of craft beer.
+
+    Returns
+    -------
+    ax : axes
+        A matplotlib axes containing the plotted data
+
+    """
+
```
wasade: extra line

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    # Gets the data vectors
+    vecs = [grouped.get_group(g)[cat].values for g in order]
+
+    # Formats the axis, if not already done
+    if 'xticklabels' not in boxplot_props:
+        boxplot_props['xticklabels'] = [g.split('(')[0] for g in order]
+    if 'show_xticks' not in boxplot_props:
+        boxplot_props['show_xticks'] = False
+    if 'show_ygrid' not in boxplot_props:
+        boxplot_props['show_ygrid'] = True
```
wasade: shouldn't this be `False`?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        When `show_stats` is True, a summary of each group will be displayed
+        along with the p values. 
+
+    Returns
+    -------
+    post_hoc : dataframe
+        `post_hoc` summarizes the results of the post-hoc test. It includes
+        statitics about each distribution, as well as the comparison matrix
+        of p-values.
+
```
wasade: extra whitespace

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    """
+
```
wasade: same here

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    # Groups the data
+    grouped = meta.groupby(group)
+
+    # Gets the order
+    if order is None:
+        order = grouped.groups.keys()
+
+    # Prealocates an output frame
```
wasade: this is just allocating

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    if show_stats:
+        stats = pd.DataFrame({'Counts': grouped[cat].count(),
+                              'Mean':  grouped[cat].mean(),
+                              'Stdv': grouped[cat].std(),
```
wasade: Stdv?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                              'Median': grouped[cat].median()})
+
+    # Preforms ad-hoc comparisons
+    comparison = {}
+
+    for pos, g1_name in enumerate(order[:-1]):
+        g1_data = grouped.get_group(g1_name)[cat]
+        compare = []
+        index = []
```
wasade: `index` appears to just be `order`

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        for id2, g2_name in enumerate(order):
```
wasade: This could be done fancier (non-blocking comment) using numpy. something like:

```python
mat = arange(1, len(order)) + arange(1, len(order))[:, newaxis]
mat -= tril(c)
for g1_pos, g2_pos in (mat > 0).nonzero():
    # do stuff
```

I think that works. Basically, the pattern here is that you want to produce some pairwise matrix from a vector. In this case, your end result is a vector, but if I'm looking at this right, its effectively a flattened version of the upper triangle. So what you really care about are getting the corresponding index positions you care about in that matrix (which is what the initial voodoo is doing). You can then index directly into `order`. The downside is this doesn't account for the `nan`, though that could be accounted for in the matrix (e.g., construct a matrix of `nan`, populate by the the imputed indices, and the serialize). 

These type of bulk operations are almost always significantly faster if done in numpy than in python too.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    post_hoc = post_hoc.reindex(order)
+
+    # Performs the multiple hypothesis correction
+    if correct is not None:
+        post_hoc = multiple_correct_post_hoc(post_hoc, order, correct)
+
+    return post_hoc
+
+
+def multiple_correct_post_hoc(raw_ph, order, alphafwer=0.05,
```
wasade: no docstring?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    raw_ps = raw_ph.loc[order, order[:-1]].values
+    ps_all = raw_ps.reshape(pos.max() + 1)
+    pos_all = pos.reshape(pos.max() + 1)
+
+    # Sorts the data
+    ps_ord = np.argsort(ps_all)
+    ps_sort = ps_all[ps_ord]
+    pos_sort = pos_all[ps_ord]
+    # Identifies the position of missing values in the sorted matrix
+    ps_nan = np.isnan(ps_sort) == False
```
wasade: `ps_nan = not np.isnan(ps_sort)`

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    """Renders a barchart
+
+    Parameters
+    ----------
+    height : array_like
+        The height of each of the bars
+    interval : float, optional
+        The spacing between the bars
+    width : float, optional
+        The width of each bars. Should be less than or equal to the interal.
```
wasade: interval

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    elinewidth : int
+        The weight in points of the errorbar on top of the plot.
+    ecapwidth : int
+        The weight in points of the line over the errorbar.
+    offset : float
+        The space on the axis for which the data should be offset. This allows
+        plotting multiple barcharts on the same axis with spacing, or plotting
+        the bar chart away from the origin.
+    p_value : float, optional
+        Default is None. When supplied, the signfigance value will be displayed
```
wasade: significance

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    elif isinstance(colormap, str):
+        colormap = segment_colormap(colormap, len(height))
+
+    if match_colors:
+        edgecolors = colormap
+    else:
+        edgecolors = np.array([[0, 0, 0]]*len(height))
+
+    # Gets an axis instance
+    if ax is None:
```
wasade: this is a duplicate of line 469

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    xleft = np.arange(0, len(height)) * interval + (interval - width)/2 + offset
+    xlims = [0, len(height)*interval]
+
+    if 'xlim' not in kwargs or kwargs['xlim'] is None:
+        kwargs['xlim'] = xlims
+
+    # Plots the errorbars
+    if errors is not None:
+        e_bars = []
+        for idx, err in enumerate(errors):
```
wasade: can `zip` all the things here instead of enumerate

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        one less p value than there are bars.
+    ax : matplotlib axis, optional
+        The axis where data should be plotted. If none, a new axis instance
+        will be created.
+    space : float
+        The y-distance between each the comparison line.
+    interval : float
+        The space between the bars
+    lowest : float, optional
+        If mutliple sets of comparison bars are being added, this allows
```
wasade: multiple

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        the user to set the position of the lowest bar in the group
+    factor : unsigned int, optional
+        The ones-place value to which values will be rounded. For example,
+        for a `factor` of 5, a value of 0.12 will be rounded to 0.15 and a
+        value of 2.7 will be rounded to 3.0.
+    label_size : unsigned int, optional
+        The font size for displaying signifigance labels
```
wasade: significance

...gonna stop mentioning this one. suggest doing a search through the code for "signi" and check that way?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        p <= 0.01: '**', p <= 0.001: '***', p <= 0.0001).
+
+    Returns
+    -------
+    lines : list
+        A list of the lines and text objects which have been plotted
+
+    """
+    # Checks the shapes of the inputs
+    if not centers.shape == tops.shape:
```
wasade: `if centers.shape != tops.shape`

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        raise ValueError('centers and tops must be the same length')
+    if not centers.shape[0] == (p_values.shape[0] + 1):
```
wasade: same general comment here...

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+    correct = np.power(10, -np.log10(fudge))
+
+    # Sets the spacing
+    if space is None:
+        space = correct/15.
+    if interval is None:
+        interval = correct/3.
+
+    # Sets the tops of hte bars
```
wasade: the

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+    # Identifies the center positions for the p text
+    p_cents = [(centers[1] - centers[0])/2. + centers[0]]
+
+    if len(centers) > 2:
+        for idx, center in enumerate(centers[2:]):
+            p_cents.append((center - centers[0])/2. + centers[0])
+    # Determines the p text
+    p_text = []
+    for p in p_values:
```
wasade: Non-blocking, just a different take on this. The special cases for `show_value` make this difficult to collapse completely unfortunately, though the logic could be split I suppose

```python
    _p_marks = ['', '+', '*', '**', '***', '****', '*****'] # note, 1 index longer than _p_thresh
    _p_thresh = array([1.0, 0.1, 0.05, 0.01, 0.001, 0.0001])
    p_text = []
    for p in p_values:
        if p < 0.01 and show_value:
            p_text.append('%1.e' % p)
        elif show_value:
            p_text.append('%1.2f' % p)
        else:
            p_text.append(_p_marks[np.searchsorted((p > _p_thresh), True)])
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    l1 = []
+    l1.append(ax.plot([centers[0]]*2, [tops[0] + space, bar_levels[-1]], 'k-'))
+    l1.append(ax.plot([centers[1]]*2, [tops[1] + space, bar_levels[0]], 'k-'))
+    l1.append(ax.plot([centers[0], centers[1]], [bar_levels[0]]*2, 'k-'))
+    l1.append(ax.text(x=p_cents[0], y=bar_levels[0] + 0.2*space, s=p_text[0],
+                      size=label_size, ha='center'))
+    lines.append(l1)
+
+    if len(centers) > 2:
+        for idx, p_cent in enumerate(p_cents[:-1]):
```
wasade: `p_cent` isn't used in the loop?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+            ln.append(ax.text(x=p_cents[idx + 1],
+                              y=bar_levels[idx + 1] + 0.2*space,
+                              s=p_text[idx+1], size=label_size, ha='center'))
+            lines.append(ln)
+
+    return lines
+
+
+def segment_colormap(cm_name, n_colors, n_pad=None, start=None):
+    """Segments a matplotlib colormap into descrete colors
```
wasade: discrete

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+    Parameters
+    ----------
+    cm_name : str
+        The name of the continous matplotlib colormap which will be segmented.
```
wasade: continuous

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        will be one more than the number of colors provided.
+    start : int, optional
+        The number of colors to skip over before display starts. By default,
+        this is `n_pad` - `n_colors`.
+
+    Returns
+    -------
+    new_map : array
+        A segmented array containing the colormap
+
```
wasade: extra whitespace

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    # Sets parameters if necessary
+    if n_pad is None:
+        n_pad = n_colors + 1
+    if start is None:
+        start = n_pad - n_colors
+
+    # Gets the colormap
+    cm = mpl.cm.get_cmap(cm_name)
+    # Sets up the new map
+    new_map = np.array([list(cm(1. * (i + start) / n_pad)) for i in
```
wasade: why cast to `list`? this is essentially an `np.array` of a `list` of `list`s, or is the return from a `Colormap` not numpy friendly?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                        xrange(n_colors)])
+
+    return new_map
+
+
+def _get_bar_height(tops, factor=5):
+    """Calculates the lowest bar height"""
+    max_hi = tops.max()
+    # Gets the correct order of magnitdue
```
wasade: magnitude

americangut/diversity_analysis.py
JWDebelius: The default for `_format_axis` is to not to show grids, since they're discouraged in almost every primer on preparing scientific figures. The lab seems to prefer grid lines on boxplots, and I'd make the case that they are help for for interpretation, so I made it default here.

americangut/diversity_analysis.py
JWDebelius: Standard devation. stdv is the function in excel.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        lowest = np.ceil(max_hi*fudge/factor)/(fudge)*factor
+    else:
+        lowest = np.ceil(max_hi*fudge*10/factor)/(fudge*10)*factor
+
+    return lowest, fudge
+
+
+def _get_p_value(sub_p, ref_group, group, p_tab_col):
+    """Determines the comparison value within a post-hoc table"""
+    if '%s vs. %s' % (ref_group, group) in sub_p['Group 2'].values:
```
wasade: this can be simplified. given that it is used in an inner loop, this method should likely be optimized some. these if/elif blocks appear to be doing repeated linear time lookups (the `foo in blah.values`) as the `.values` is likely an `array` or `list`. from looking at the use of `_get_p_value`, it looks like `sub_p` varies outside of the forloop that this method is being called in. That suggests there may be some caching benefit. For instance, in the calling method, outside of the nested forloop this method is called in, something like this could be done:

```python
sub_p_lookups = {k: set(sub_p[k].values) for k in ('Group 1', 'Group 2')}
```

Then `_get_p_value` can be reduced some to:
```python
query_fwd = '%s vs. %s' % (ref_group, group)
query_rev = '%s vs. %s' % (group, ref_group)

for query, group in itertools.product([query_fwd, query_rev], ['Group 1', 'Group 2']):
    if query in sub_p_lookups[group]:
        return sub_p.loc[sub_p[group] == query, p_tab_col].values[0]
raise ValueError('%s is not a defined group' % query_fwd)
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    else:
+        raise ValueError('%s vs. %s is not a defined group'
+                         % (ref_group, group))
+
+    return p_value
+
+
+def _correct_p_value(tail, p_value, ref_val, current_val):
+    """Determines if p-values should be tail corrected"""
+    if tail and ref_val > current_val:
```
wasade: `return 1.0 if tail and ref_val > current_val else p_value`

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+def get_distance_vectors(dm, df, group, order=None):
+    """Extracts the distance information for all samples in a group
+
+    Parameters
+    ----------
+    dm : skbio DistanceMatrix
+        A distance matrix object with the samples corresponding to those in
+        the data frame with the metadata.
+    df : pandas DataFrame
+        A dataframe containing the metadata associated withed with the object
```
wasade: with

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    group : str
+        the metadata category being interograted, used to group the data
```
wasade: interrogated

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    order : None, list
+        Default is None. The order of groups in group.
```
wasade: What does an order of `None` mean?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    if group not in df:
+        raise ValueError('%s is not a metadata category.' % group)
+
+    # Gets the sample ids associated with the group
+    if order is None:
+        order = list(df.groupby(group).groups)
+    ordered_ids = {o: df.groupby(group).groups[o] for o in order}
+
+    # Gets the data
+    dist_ids = np.array(list(dm.ids))
```
wasade: cast to `list` isn't necessary

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    between = []
+    b_dist = []
+
+    # Loops through the groups
+    for id1, o1 in enumerate(order):
+        # Gets the intragroup distance
+        within.append(o1)
+        w_dist.append(dm.filter(ordered_ids[o1]).condensed_form())
+        for o2 in order[(id1+1):]:
+            loc1 = np.array([dist_ids == i for i in ordered_ids[o1]]).any(0)
```
wasade: I _think_ these can be rewritten as:

```python
loc1 = (ordered_ids[o1] == dist_ids).any()
loc2 = (ordered_ids[o2] == dist_ids).any()
```

But I don't understand it as `loc1` and `loc2` will be boolean yet they appear to be used as indices?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+    # Gets the data
+    dist_ids = np.array(list(dm.ids))
+    dist_data = dm.data
+
+    # Prealocates objects for return
+    within = []
+    w_dist = []
+    between = []
+    b_dist = []
```
wasade: It would take a bit of thought, but it should be possible to determine the size of these objects. What that means is you could preallocate (e.g., `b_dist = zeros(the_number_of_resulting_elements, dtype=float)`) and then just fill the vectors. It would rework this some, but if this code is slow (suggest profiling first), then it would be one way to definitely speed it up. The `append` calls in the inner loop, if done a lot, will get expensive particularly if the resulting `list`s are large in size due to how Python allocates memory

americangut/diversity_analysis.py
wasade: Never used stdv in excel, but I actively try to avoid that piece of
software if possible. Flagged because I've never seen standard deviation
referred to in python (or other packages) as that. no need to change, just
curious

americangut/diversity_analysis.py
wasade: flagged due to the inconsistency of the logic. If `show_ygrid` is not in
your options, then the default is to show the grid. This is the exact
opposite logic to the prior if statement.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    ----------
+    dm : skbio DistanceMatrix
+        A distance matrix object with the samples corresponding to those in
+        the data frame with the metadata.
+    meta : pandas dataframe
+        The metadata associated with the samples in the DataFrame. The samples
+        in the DataFrame may be a superset of the samples in the distance
+        matrix, `dm`.
+    group : str
+        the metadata category being interograted, used to group the data
```
wasade: interrogated

again, copy and paste is dangerous but for more than just spelling

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        Default is True. Adds horizonal lines at each major y-tick.
+    title: str, optional
+        The title to be placed on the graph.
+    ylims : list
+        The limits for the y-axis.
+    ylabel : str
+        The label text for the y-axis. Every time you leave off appropriate
+        labels and units, a science grad student grading lab reports cries
+        another bitter tear into their bottle of craft beer.
+
```
wasade: extra whitespace

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        get_distance_vectors(dm, meta, group, order)
+
+    bar_counter = 0
+
+    # Loops through the data to make boxplots compared to the reference group
+    for id1, ref_group in enumerate(ref_groups):
+        # Determines the offset
+        offset = (id1 * len(order) + bar_counter)*width
+        # Determines the position of the reference group
+        ref_loc = order.index(ref_group)
```
wasade: this is a linear search, why not just iterate over `order.index`?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+            sub_p = pd.concat([p_table.loc[p_table['Group 1'] == '%s vs. %s'
+                                           % (ref_group, ref_group)],
+                               p_table.loc[p_table['Group 2'] == '%s vs. %s'
+                                           % (ref_group, ref_group)]])
+
+        for id2, group in enumerate(order):
+            if group == ref_group:
+                continue
+            # Gets the distance vector
+            dist_bar[id2] = b_dist[b_groups.index({ref_group, group})].mean()
```
wasade: those `.index` calls are linear searches. It may be worth investigating a more efficient means of retrieval

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                            match_colors=match_colors,
+                            elinewidth=elinewidth,
+                            ecapwidth=ecapwidth)
+        # Watches the xlimits
+        if id1 == 0:
+            xlim = [xpos[0] - interval*0.75]
+
+        # Gets the critical lines to display on the figure
+        if lowest is None:
+            lowest, __ = _get_bar_height(dist_bar+dist_std)
```
wasade: generally its just a single `_` that is used

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                                       show_value=show_p_value)
+            bars2 = add_comparison_bars(xpos[:(ref_loc+1)][::-1],
+                                        dist_bar[:(ref_loc+1)][::-1] +
+                                        dist_std[:(ref_loc+1)][::-1],
+                                        p_values[:(ref_loc)][::-1],
+                                        ax, lowest=lowest,
+                                        label_size=label_size,
+                                        show_value=show_p_value)
+            bars.extend(bars2)
+
```
wasade: it's valid for `p_values` to be `None`?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                            ecapwidth=ecapwidth)
+        # Watches the xlimits
+        if id1 == 0:
+            xlim = [xpos[0] - interval*0.75]
+
+        # Gets the critical lines to display on the figure
+        if lowest is None:
+            lowest, __ = _get_bar_height(dist_bar+dist_std)
+        if p_values is not None and (ref_loc == 0 or len(dist_bar) == 2):
+            bars = add_comparison_bars(xpos,
```
wasade: this block likely be collapsed as there is a lot of repetition

wasade:
I gotta take a break from this. Will try to get back to this soon -- halfway through.

americangut/diversity_analysis.py
JWDebelius: That gives me a single bool value. The version I'm using is building a boolian index. I'm using it to filter the distance matrix. There's probably a better way to do thing by knowing the index positions in the distance matrix. I do the list comprehension to build up two one-dimensional vectors.

JWDebelius: That gives me a single boolian value. 

The list comprehension gives me a boolian array I can use to filter the distance matrix so I get an asymmetrical matrix. There is probably a fancier way to do it which involves understanding the way the condensed return indexes.

americangut/diversity_analysis.py
JWDebelius: This returns a single boolian value, and `ps_nan` needs to be an array.

wasade: Sorry, np.logical_not

americangut/diversity_analysis.py
wasade: Thats what I'm confused about, usually indices are integers, not bool, as
bool can only index 2 positions

JWDebelius: But you can make a numpy array of boolian vaues that you can use an array of index values. It's like a binary mask?

americangut/diversity_analysis.py
JWDebelius: If a `p_table` wasn't passed, `p_values` can be `None`.

In theory it's there for a quick, dirty approach to look at the data without spending time on make_distance_boxplots.

americangut/diversity_analysis.py
wasade: Is `ordered_ids[o1]` itself a 2-D matrix? If so, then this changes everything, however if that is the case, then please make this clearer. 

Note that if it is a 2-D matrix, then you can still do the following which will be more efficient than the list comprehension 

```python
09:29:21 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt/Downloads
> a = arange(25).reshape(5, 5)

09:29:29 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt/Downloads
> b = array([15, 16, 17, 18, 19])

09:29:31 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt/Downloads
> a == b
Out[37]:
array([[False, False, False, False, False],
       [False, False, False, False, False],
       [False, False, False, False, False],
       [ True,  True,  True,  True,  True],
       [False, False, False, False, False]], dtype=bool)

09:29:36 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt/Downloads
> (a == b).any(1)
Out[38]: array([False, False, False,  True, False], dtype=bool)
```

americangut/diversity_analysis.py
wasade: can you add something like:

```python
else:
    # p_values is None in this case, just a shortcut for by-passing make_distance_boxplots
    pass
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                                        label_size=label_size,
+                                        show_value=show_p_value)
+            bars.extend(bars2)
+
+        # Advances the bar count
+        bar_counter = bar_counter + 1
+    xlim.append(xpos[-1] + interval*0.75)
+
+    # Sets up axis formatting defaults, if not supplied
+    if 'show_frame' not in kwargs:
```
wasade: Just a note, and this is minor and non-blocking, but you can reduce the lines of code (if you haven't seen it, commonly referred to as LOC), by doing something like:

```python
kwargs['show_frame'] = kwargs.get('show_frame', False) 
```

LOC comes up in formal measures of code complexity. The basic principle is that, more lines of code, means more for the developer to digest increasing the difficulty of fully understanding a method. Its somewhat analogous to a run-on sentence vs. a short succinct one where the latter is generally easier and quicker for the reader to understand.

wasade: Actually, going a step further, there appears to be a lot of replication in what's going on with the `kwarg`s, making this essentially boiler plate code. One thing that could be done would be to decompose the kwargs stuff into a single function whose responsibility is setting plotting defaults. Something like:

```python
def _plot_defaults(kwargs):
    # ...
    kwargs['show_frame'] = kwargs.get('show_frame', False)
    # ...
```

The method could be used as:

```python
_plot_defaults(kwargs)
```

And the method does not need to return as modifications to a `dict` are in place. This would likely greatly reduce code. However, it of course assumes that the defaults are common across plot types. If that is not the case, then this strategy gets much more complex and likely not worth doing.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        the taxonomy data summarized to the specified level. If data is missing
+        for a particular level, the string will specify the last known string.
+        (i.e. k__kingdom;p__phylum;c__pclass;o__porder;f__family;g__;s__
+        outputs with a genus and species value "f. family").
+
+    levels: list
+        the name of the taxonomic levels corresponding to columns in the table.
+    """
+
+    if level not in {2, 3, 4, 5, 6, 7}:
```
wasade: `if not (2 <= level <= 7)`

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        raise ValueError('%r is not a supported taxonomic level.' % level)
+
+    # Prealocates a holding object
+    splits = np.zeros((1, (level)))
```
wasade: `splits = np.zeros((1, level))`

the extra parentheses around level don't do anything

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+    # Sets up the names of phylogenetic levels
+    levels = ['kingdom', 'phylum', 'p_class', 'p_order', 'family', 'genus',
```
wasade: why `p_class` and `p_order`?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                    clean.append(last)
+                elif '[' in val:
+                    v_corr = 'cont. %s' % val.replace('[', '').replace(']', '')
+                    clean.append(v_corr)
+                    last = '%s. %s' % (t.split('__')[0], v_corr)
+                else:
+                    clean.append(val)
+                    last = t.replace('__', '. ')
+        # Adds the information to the taxa watch
+        splits = np.vstack((splits, np.array(clean)))
```
wasade: the "preallocating" done above is not being taken advantage of. What you want to do is preallocate, and then fill the content of that matrix, not continue to grow it.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        if taxon == 'Unassigned':
+            clean = ['Unassigned']*level
+        else:
+            # Splits the greengenes string
+            rough = [v.strip() for v in taxon.split(';')]
+            # Watches the last space in the data
+            try:
+                last = rough[level - 1].replace('__', '. ')
+            except:
+                raise ValueError('There are %i levels in your taxa string and'
```
wasade: this should never happen as greengenes strings are assured to be 7 level, and there is a bounds check on `level`

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                                 ' you would like to look at %i levels.'
+                                 % (len(rough), level))
+            clean = []
+
+            # Cleans the taxa
+            for t in rough[:(level)]:
```
wasade: parentheses aren't doing anythign

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    # Loops through the list of taxa
+    for taxon in taxa:
+        if taxon == 'Unassigned':
+            clean = ['Unassigned']*level
+        else:
+            # Splits the greengenes string
+            rough = [v.strip() for v in taxon.split(';')]
+            # Watches the last space in the data
+            try:
+                last = rough[level - 1].replace('__', '. ')
```
wasade: why is the last level special cased?

wasade: i see now..

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+            # Cleans the taxa
+            for t in rough[:(level)]:
+                val = t.split('__')[1].replace('_', ' ').replace('-', ' ')
+                if val == '':
+                    clean.append(last)
+                elif '[' in val:
+                    v_corr = 'cont. %s' % val.replace('[', '').replace(']', '')
+                    clean.append(v_corr)
+                    last = '%s. %s' % (t.split('__')[0], v_corr)
```
wasade: only need to do the split on `__` once. suggest setting up the rank, taxon name split first thing in the for loop

americangut/diversity_analysis.py
wasade: except that last is almost always going to be wiped out by line 1132. why not do the forloop below processing below first, and then determine what is "last"?

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                else:
+                    clean.append(val)
+                    last = t.replace('__', '. ')
+        # Adds the information to the taxa watch
+        splits = np.vstack((splits, np.array(clean)))
+
+    return splits[1:], levels[:level]
```
wasade: `splits` shouldn't need to be sliced if it is preallocated to the correct shape

americangut/diversity_analysis.py
wasade: this is preallocating a single 7 element vector. based on the code below, you want to instead do:

```python
splits = np.zeros((len(taxa), level), dtype=object)
# ...

for i, taxon in enumerate(taxa):
    # ...
   
    clean = zeros(level, dtype=object)
    for j, t in enumerate(rough[:level]):
        # ...
        clean[j] = foo  
    splits[idx, :] = clean
```

americangut/diversity_analysis.py
JWDebelius: I tried to set as many common defaults as possible in `_format_axis`. The defaults that appear in functions are usually because they make sense for that particular plot instance.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        A numpy array with the the ratio of the data values to the reference
+        group.
+
+    """
+    if ref_pos is None:
+        ref = data.mean(1)
+    else:
+        ref = data[:, ref_pos]
+
+    ref = ((ref*np.ones((1, data.shape[0]))) *
```
wasade: is this the same as:

```python
ref = ref * np.ones(data.shape[::-1])
```

americangut/diversity_analysis.py
JWDebelius: because class and order are python keywords, and calling `df.class` of `df.order` is already defined.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+           np.ones((data.shape[1], 1))).transpose()
+
+    if log is None:
```
wasade:
```python
ratio = data / ref
if log is not None:
    ratio = np.log(ratio) / np.log(log)
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+
+def heatmap(data, ax=None,  cmap='RdBu_r', clims=None, cbar_size=11, **kwargs):
+    """Wraps matplotlib's heatmap and formats the colorbar
+
+    Parameters
+    ----------
+    data : array
+        The data to be plotted in the heat map
+   ax : matplotlib axis, optional
```
wasade: needs a space

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    -------
+    ax: matplotlib axis
+        The axis with the data plotted.
+    cbar : matplotlib colorbar
+        The colorbar instance on the plot
+
+    """
+    # Checks the shape of the data is sane
+    mat_shape = data.shape
+    if ('xticklabels' in kwargs and kwargs['xticklabels'] is not None and not
```
wasade:
```python
if kwargs.get('xticklabels', None) is not None and len(kwargs.get('xticklabels', [])) != mat_shape[0]:
    # ...
```

Note, "xticklabels" is described as optional, but the conditional (as it is written in the PR) requires it

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        len(kwargs['xticklabels']) == mat_shape[1]):
+        raise ValueError('There must be a label for each column in '
+                         'data.')
+
+    if ('yticklabels' in kwargs and kwargs['yticklabels'] is not None and not
```
wasade: same revisions here.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    if 'xticks' not in kwargs:
+        kwargs['xticks'] = np.arange(0, mat_shape[1]) + 0.5
+    if 'yticks' not in kwargs:
+        kwargs['yticks'] = np.arange(0, mat_shape[0]) + 0.5
+
+    # Formats the axis
+    _format_axis(ax, **kwargs)
+
+    # Formats the colorbar ylabels
+    new_labels = ['%s' % t.get_text() for t in cbar.ax.get_yticklabels()]
```
wasade: It doesn't look like the string formatting is doing anything...

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        The column name from which signifigance values should be drawn. It is
+        recommended that 'Bonferroni_P' or 'FDR_P' be used.
+    p_crit : float, optional
+        The critical p value. Comparison p values must be less than this
+        for the comparisons to be displayed on the heatmap.
+    ref : str, optional
+        The name of the column which should serve as the refence for the ratio
+        heatmap. If none is supplied, the arethmatic mean will be used.
+    ratio_base : float, optional
+        The logarthemic base for the ratio heatmap. If `None`, then no
```
wasade: logarithmic

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        logarthem will be taken.
```
wasade: logarithm

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        Default is 12. The size of the x tick labels
+    yfont_size : int, optional
+        Default is 12. The size of the y tick labels
+
+    Returns
+    -------
+    ax1, ax2 : matplotlib axes
+        The matplotlib axis instances with the raw and ratio data, respectively
+    cbar1, cbar2 : matplotlib colorbars
+        The colobar instances for the raw and ratio data, respectively.
```
wasade: colorbar

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    sig = gs.loc[crit]
+
+    # Adds taxa to the signifigant dataframe
+    taxa, levels = split_taxa(sig.taxonomy.values, 7)
+    sig = sig.join(pd.DataFrame(taxa, columns=levels, index=sig.index))
+
+    # Creates an index column
+    if label == 'INDEX':
+        sig['label'] = sig['genus'] + ' (' + \
+            sig['OTU'].apply(lambda x: str((x))) + ')'
```
wasade: extra set of parens around `x`

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    else:
+        sig['label'] = sig[label]
+
+    # If approprate orders the tables taxonomically
+    if sort_by_taxa:
+        sig = sig.sort(['phylum', 'p_class', 'p_order', 'family',
```
wasade: same comment here about "p_class" etc

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+                         cmap=cmap2,
+                         clims=clims2,
+                         cbar_size=cbar_size,
+                         xticklabels=g_labels,
+                         yticklabels=['']*len(sig['label'].values),
+                         **kwargs)
+    ax1.set_xlabel('Raw Data', size=15)
+    ax2.set_xlabel('Ratio Data', size=15)
+
+    if consistant_size:
```
wasade: consistent_size

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        ax1.set_position(a1p)
+        ax2.set_position(a2p)
+        cbar1.ax.set_position(c1p)
+        cbar2.ax.set_position(c2p)
+
+    return [ax1, ax2], [cbar1, cbar2]
+
+
+def _format_axis(ax, **kwargs):
+    """Beautifies the axis
```
wasade: that's extremely general

americangut/diversity_analysis.py
JWDebelius: You can pass in a taxa table with a taxonomy string. This error is here because I was passing a genus or phylum table, and forgetting to reset level.

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+            'yticks': None,
+            'yticklabels': "text",
+            'ytick_format': '%s',
+            'ylabel': '',
+            'yfont_angle': 0,
+            'yfont_align': 'right',
+            'ylabel_size': 15,
+            'ytick_size': 12}
+
+    for key, val in kwargs.iteritems():
```
wasade: fairly certain you can just do:

```python
kwds.update(kwargs)
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+
+    # Sets axis ticks
+    if kwds['xticks'] is not None:
+        ax.set_xticks(kwds['xticks'])
+
+    if kwds['yticks'] is not None:
+        ax.set_yticks(kwds['yticks'])
+
+    # Checks the max and min values values
+    if kwds['xlim'] is not None and not len(kwds['xlim']) == 2:
```
wasade: please remember to use the `!=` operator...

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+    # Sets axis limits
+    if kwds['xlim'] is not None:
+        ax.set_xlim(kwds['xlim'])
+    if kwds['ylim'] is not None:
+        ax.set_ylim(kwds['ylim'])
+
+    # Formats the tick labels
+    if kwds['xticklabels'] is None:
+        xtls = ['']*len(ax.get_xticks())
+    elif (isinstance(kwds['xticklabels'], str) and
```
wasade: what else could `xticklabels` be? if None, just test explicitly:

```python
elif kwds['xticklabels'] is not None and kwds['xticklabels'].lower() == 'text':
```

americangut/diversity_analysis.py
```diff
@@ -0,0 +1,1729 @@
...
+        raise ValueError('There must be a label for each xtick.')
+    else:
+        xtls = kwds['xticklabels']
+
+    ax.set_xticklabels(xtls,
+                       ha=kwds['xfont_align'],
+                       rotation=kwds['xfont_angle'],
+                       size=kwds['xtick_size'])
+
+    if kwds['yticklabels'] is None:
```
wasade: conditional block appears to be essentially the same as the one above, please decompose (could just do a local function or something)

wasade: so much of this code is effectively replicated over each axis. i much of this method could be reduced substantially

wasade: it's non-blocking... but reducing this method would be nice

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+
+        if len(power_shape) == 1:
+            if not count_shape[0] == power_shape[0]:
+                raise ValueError('There must be a sample count for each '
+                                 'power.')
+        elif not count_shape[0] == power_shape[1]:
+            raise ValueError('There must be a sample count for each power.')
+
+    # Prealocates the output arrays
+    effect_means = np.zeros((num_powers))
```
wasade: extra set of parentheses

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+        An array with the mean effects. The output of `collate_effect_size`.
+    a_eff_bounds, b_eff_bounds : list of arrays
+         An array with the mean bound. The confidence interval for the mean
+         is given by [eff_mean - eff_bound, eff_mean + eff_bound].
+         This is an output of `collate_effect_size`.
+
+    Returns
+    -------
+    table : str
+        An html-formmated table which can be rendered in IPython.
```
wasade: html-formatted

americangut/diversity_analysis.py
JWDebelius: This converts the labels from a text object to a string that can be operated on. Removing the `$` lets the font for the color bar match the rest of the fonts in the figure if using LaTeX for font rendering.

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+    # Loops through each row
+    for idx in order:
+        # Gets the effect sizes
+        cat = fecal_cats[idx]
+        a_eff_m = a_eff_means[idx]
+        a_eff_b = a_eff_bounds[idx]
+        b_eff_m = b_eff_means[idx]
+        b_eff_b = b_eff_bounds[idx]
+        # Gets the fitted values
+        a_eff_fit = np.ceil(ft.solve_power(a_eff_m, nobs=None, power=0.8,
```
wasade: why is solving happening here? i thought this method was just responsible for the construction of the html table?

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+               '\t\t          ">',
+               '\t\t\t&plusmn;',
+               '\t\t</td>',
+               '\t\t<td style="border-top:hidden;',
+               '\t\t           border-bottom:hidden;',
+               '\t\t           border-left: hidden;',
+               '\t\t           border-bottom: hidden;',
+               '\t\t           text-align:right',
+               '\t\t          ">',
+               '\t\t\t%i' % b_eff_rnd,
```
wasade: just fyi, the vast majority, if not all, of the tab characters aren't doing anything...

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+               '\t\t</td>',
+               ]
+        table.append('\n'.join(row))
+    table.append('</table>')
+
+    return '\n'.join(table)
+
+
+def plot_effects(effect_means, effect_bounds, labels, sample_counts, **kwargs):
+
```
wasade: extra whitespace

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+    Returns
+    -------
+    fig : figure
+        a figure with the power curves plotted.
+
+    Other parameters
+    ----------------
+    leg_offset : tuple
+        Changes the legend position.
+    tick_size : usigned int
```
wasade: unsigned

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+            'colormap': None,
+            'grid': True,
+            'title': '',
+            'show_bound': True,
+            'leg_offset': None,
+            'tick_size': 12,
+            'label_size': 15,
+            'title_size': 18,
+            'legend_size': 11}
+    for key, value in viewitems(kwargs):
```
wasade: `kwds.update(kwargs)`

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+
+    # Checks the effect, bound, and mean argument is sane
+    mean_shape = effect_means.shape
+    if effect_bounds is None:
+        kwds['show_bound'] = False
+        effect_bounds = np.zeros(mean_shape)
+    bound_shape = effect_bounds.shape
+    label_shape = labels.shape
+
+    if not len(mean_shape) == 1:
```
wasade: please use the correct operator

americangut/diversity_analysis.py
JWDebelius: It's probably an array or list if you're setting new labels.
This block is used to use the x-position as the label, and then make the text look right.

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+        pwr_mean, pwr_lower, pwr_upper : ndarray
+            The mean, upper and lower bounds for the power array.
+
+    """
+
+    pwr_mean = np.nanmean(power, 0)
+    pwr_bound = confidence_bound(power, axis=0)
+
+    # Prevents the confidence interval of being less than 0 or more than 1
+    pwr_lower = np.vstack((pwr_mean - pwr_bound,
```
wasade: isn't this the same as:

```python
pwr_lower = (pwr_mean - pwr_bound).max(0)
```?

I don't understand the need for stacking a zero'd matrix that is the same shape as `pwr_mean`...

File: americangut/power_plots.py [hunk 87]
```diff
@@ -0,0 +1,581 @@
...
+def add_average_trace(fig, power, counts, labels, **kwargs):
+    """Adds a trace of averaged points
+
+    This can be used with power curves which are too steep to use to estimate
+    the effect size.
+
+    Parameters
+    ----------
+    fig : figure
+        A matplitlib figure instance generated by `plot_effects` which has
```
wasade: matplotlib

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+    # Sets up the keyword arguements
+    kwds = {'figure_size': (7., 3.5),
+            'axis_size': (0.6, 0.6, 4., 2.5),
+            'legend_pad': (5., 0.6, 2., 2.5),
+            'legend_font_size': 11,
+            'legend_position': (1.05, 0.95),
+            'color': [0.5, 0.5, 0.5]}
+
+    # Updates the keyword arguments
+    for kwd, val in viewitems(kwargs):
```
wasade: `kwds.update(kwargs)`

americangut/power_plots.py
```diff
@@ -0,0 +1,581 @@
...
+
+    # Adds a new axis to pad for the legend display
+    pad_ax = fig.add_axes(pad_dims)
+    pad_ax.set_visible(False)
+
+    # Hides the current axis legend so a new legend showing body site can be
+    # added.
+    axes[2].get_legend().set_visible(False)
+
+    # Adjusts the position of the filte vector so the bodysite line will appear
```
wasade: filter

ipynb/index.ipynb
```diff
@@ -16,7 +17,11 @@
       "* Primary processing pipeline [Notebook](module2_v1.0.ipynb)\n",
       "* Alpha diversity analysis [Notebook](Alpha%20diversity%20notebook.ipynb)\n",
       "* Shared OTUs [Notebook](shared_otus.ipynb)\n",
+<<<<<<< HEAD
```
wasade: please make sure conflicts are resolved apprioriately

wasade:
i believe i'm done with the static analysis

ElDeveloper:
I cannot review Alph Diversity.ipynb the web app is not displaying this file. So, I am not going to review that file, sorry.

americangut/diversity_analysis.py
wasade: i dont see that happening anywhere in the code, and they are not reserved strings

americangut/diversity_analysis.py
wasade: `new_labels = [str(t.get_text()) for t in cbar.ax.get_yticklabels()]` then please

wasade:
@JWDebelius, can you please fix that notebook?

americangut/diversity_analysis.py
wasade: okay. `elif isinstance(kwds['xticklabels'], (str, unicode)) and kwds['xticklabels'].lower() == 'text':`

ElDeveloper:
I don't think this is a problem with the notebook itself, but with the
fact that there's a maximum number of diffs that GitHub will render for
a given PR. In this case I think that maximum was exceeded. Really what
would have needed to happen is for this PR to have been split in more
manageable groups of changesets. In addition that would have also made
the reviewing process easier; it is usually advised to **not** have
large PRs.

JWDebelius:
Both the ways i've tried to view in [nbviewer](http://nbviewer.ipython.org/github/JWDebelius/American-Gut/blob/statistical_power/ipynb/Alpha%20diversity%20notebook.ipynb) have rendered fine.

@ElDeveloper: This was originally split into two pull requests (see #125). I was asked to combine them.

wasade:
Both were still extremely large, and I did not know that the inclusion of
the necessary code for #127 to be complete would double its size

ElDeveloper:
What I mean is that github is not rendering the notebook and thus I
cannot comment on the file.

americangut/power_plots.py
JWDebelius: Because you need to convert the effect size into a value. So, you have to solve the power.

americangut/power_plots.py
JWDebelius: That would allow negative values. Power must be between 0 and 1. This and the subsequent stacking line force the calculated bounds to also be between those values.

JWDebelius:
Would it work to comment on #125 and then include the changes here? I realize that may not be an appropriate use of the technology, but it might get around the size problem.

americangut/power_plots.py
wasade: why is this happening in a function that is responsible for producing a visual representation of the data? it is a very good idea to separate presentation from business logic...

americangut/power_plots.py
wasade: Right, but you do not need to construct a of the same shape as `pwr_mean`, full of zeros, in order to get a single zero.

```python
pwr_lower = zeros(pwr_mean.shape[1] + 1)
pwr_lower[1:] = (pwr_mean - pwr_bound).max(0)
```

JWDebelius: `pwr_lower` returns a 1d array, the same shape as `pwr_mean`, where each position in the array can take on a value [0, 1]. Vertical stacking means that I take the max of each column, not returning a single value.
The other way I can see to do this would be a list comprehension, and that would be a lot slower, from my understanding.

ElDeveloper:
The problem is that once the modifications are made in one place,
there's not a good way to verify/comment on subsequent changes.

americangut/power_plots.py
wasade: Given what the code is doing:

```python
> a
Out[4]:
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14],
       [15, 16, 17, 18, 19]])

14:50:11 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> b = vstack([a, zeros(a.shape)])

14:50:24 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> b
Out[6]:
array([[  0.,   1.,   2.,   3.,   4.],
       [  5.,   6.,   7.,   8.,   9.],
       [ 10.,  11.,  12.,  13.,  14.],
       [ 15.,  16.,  17.,  18.,  19.],
       [  0.,   0.,   0.,   0.,   0.],
       [  0.,   0.,   0.,   0.,   0.],
       [  0.,   0.,   0.,   0.,   0.],
       [  0.,   0.,   0.,   0.,   0.]])

14:50:25 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> b.max(0)
Out[7]: array([ 15.,  16.,  17.,  18.,  19.])
```

It looks like you can just do exactly what I put in the first comment and not even do the call to zeros...

JWDebelius: There is no reason to add zeros for a two-dimensional array, because you're taking the max over several values.
`eff_means` and `eff_bounds` are one-dimensional arrays, so their difference will be a  one-dimensional array. If `a` is a one-dimensional array, `a.max(0) = a`.

wasade: If there is no reason to add the zeros, why is your code adding the zeros? Note: this is why I'm suggesting to NOT add the zeros.

The code I executed in the python session is effectively the same as what you're doing, and I'm offering a better, more succinct way of doing it. 

I strongly encourage you to play around with these things in an interactive interpreter session.

JWDebelius: I've actually spent a lot of time looking at the problem. Your version is more succinct, but it doesn't solve the problem of constraining a one-dimensional array.
It's not necessary to add the zeros *when* you're dealing with a 2-dimensional array. The arrays I pass are 1-dimensional.

wasade: Then it doesn't look like the `max(0)` is doing anything...

```python
15:49:51 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> a = arange(5)

15:50:04 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> vstack([a, zeros(a.shape)])
Out[7]:
array([[ 0.,  1.,  2.,  3.,  4.],
       [ 0.,  0.,  0.,  0.,  0.]])

15:50:09 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> vstack([a, zeros(a.shape)]).max(0)
Out[8]: array([ 0.,  1.,  2.,  3.,  4.])
```

Can you please provide an example of what these data actually look like?

JWDebelius: This is masking negative values and constraining the data between 0 and 1. The math works on the assumption that power can only take on values [0, 1], since it's a probability. 
The complete solution would look like this:
```vstack(vstack((a, zeros(a.shape)).max(0), ones(a.shape)).min(0)```

But, we're assuming that `p_mean` <= 1, so `p_mean - p_bound` <= 1. So, there's no point in checking an upper bound.

Cases where it matter are ones like below:
```
> a = arrange(-0.2, 0.1, 0.3)
> a
array([-0.2, -0.1,  0. ,  0.1,  0.2])
>  vstack((a, zeros(a.shape)))
array([[-0.2, -0.1,  0. ,  0.1,  0.2],
       [ 0. ,  0. ,  0. ,  0. ,  0. ]])
>  vstack((a, zeros(a.shape))).max(0)
array([0., 0., 0., 0.1, 0.2])
```

Comment (wasade) [inline americangut/power_plots.py]: that code doesn't work...

So, it then looks like what you want is:

```python
16:10:08 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> a = arange(-0.2, 0.3, 0.1)

16:11:41 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> where(a < 0, 0, a)
Out[23]: array([ 0. ,  0. ,  0. ,  0.1,  0.2])

16:11:42 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> b = arange(0.8, 1.3, 0.1)

16:11:52 (mcdonadt@rl1-1-222-132-dhcp):/Users/mcdonadt/ResearchWork/software/American-Gut/md
> where(b > 1, 1, b)
Out[25]: array([ 0.8,  0.9,  1. ,  1. ,  1. ])
```

JWDebelius:
@wasade, I've updated the code as per your suggestions and added the new text to the notebook.

americangut/diversity_analysis.py
wasade: There are still a few miss spellings of significance

americangut/power_plots.py
```diff
@@ -0,0 +1,570 @@
...
+            The mean, upper and lower bounds for the power array.
+
+    """
+
+    pwr_mean = np.nanmean(power, 0)
+    pwr_bound = confidence_bound(power, axis=0)
+
+    # Prevents the confidence interval of being less than 0 or more than 1
+    pwr_lower = pwr_mean - pwr_bound
+    np.where(pwr_lower < 0, 0, pwr_lower)
```
wasade: This is not an inplace operation, so this doesn't have an effect. What you want here is `pwr_lower = np.where(pwr_lower < 0, 0, pwr_lower)`. 

Have the changes made on the last round of commits been verified they work as expected?

wasade:
Still a few typos I saw with 'significance', and the `trace_bounds` method has some method calls that are currently not doing anything. Going over the rendered notebook now

wasade:
@ElDeveloper, were you able to go over the other notebook by chance?

ElDeveloper:
No, I wasn't able to go over the other notebook.

wasade:
Just on a quick pass through, the rendered notebook still has grammar and spelling issues to be resolved. Where the suggestions that @cuttlefishh and @EmbrietteH made to the google doc merged in?

wasade:
Possible to do so?

JWDebelius:
I tested the code by running it through both notebooks, and the test suite.
The suggestions were merged in last night. Will run through spell check again.

wasade:
From what I'm seeing, the notebook text does not appear to be in sync with
the google doc. I mashed refresh a few times, so it shouldn't be a cached
copy

JWDebelius:
That is the updated notebook and text. What I'm seeing lines up with the
google doc, when I click both links.

ElDeveloper:
This is the reason why I can't comment on the notebook:
https://github.com/biocore/American-Gut/pull/127/files#diff-c6b18319a5fc61e394924c6a0a3ff2d3

GitHub will not show the interface for me to make comments on it.

wasade:
@JWDebelius, please see the screenshots below. There are others as well

@ElDeveloper, able to comment separate from inline on github? It's not ideal, but we need to review the material somehow

screen shot 2015-03-05 at 5 26 48 pm
screen shot 2015-03-05 at 5 27 00 pm

JWDebelius:
I've updated. I can ask someone else to check through spelling again before you give another pass.

Also, I processed the data for the alpha diversity on a pre-computed tables, since the current maps are missing preceding zeros. This skips the download step, since the notebook requires the github directory, but assumes the kernel is running in the ipnyb directory.

wasade:
Thanks. I'm going to do one more pass through today. I think we're nearly or already there, so it should go quick. And its not like we can't issue future PRs to update if necessary anyway. Thank you for all the hard work on this!!

JWDebelius:
I asked an extra person to look through it. I thought they'd be done tonight, but they sent me a couple more edits that I'm working on now. Sorry for the lengthy process.

JWDebelius:
Okay, I got the new stuff in, too. Should be good.

wasade:
Sorry, didn't see the comment until now. Reviewing

wasade:
I think this is good, thanks @JWDebelius!!!!

PR #129, opened 2015-02-24
Added sample/accession mappings for r1-14

wasade:
Constructed with:

```python
from json import dumps
sample_to_accession = dumps({l.split('\t')[0]:f.split('.')[0] for f in glob.glob('ERP*.txt') for l in open(f).readlines()[1:]})
accession_to_sample = dumps({f.split('.')[0]:[l.split('\t')[0] for l in open(f).readlines()[1:]] for f in glob.glob('ERP*.txt')})
```

cc @madprime


adamrp:
I think this looks good :+1:

wasade:
Able to merge? I don't think this warrants dual review IMO

adamrp:
I'm checking one more thing actually; not yet finding barcode 000011080 in accession ERP005651 (this is the first entry in accession_to_sample.json)

adamrp:
I think I was just missing it on their interface (which for some reason does not permit sorting by name). Just spot-checked several barcodes that are on their list against the lists here, and they all check out, so I'm going to merge this.

wasade:
Thanks!

PR #132, opened 2015-03-09
Subset and failed

### LINKED ISSUE #130: Zeros truncated in AG data mapping file

wasade:
@JWDebelius @ElDeveloper, I believe this fixes #130.

`get_failed.py` gets us info on what samples need to be rerun.


ElDeveloper:
:+1:

PR #136, opened 2015-03-20
Updating shared_otus notebook...

### LINKED ISSUE #135: The shared_otus ipynb tries to download a file that no longer exists
`AG_100nt_even10k.biom.gz -> AG_100nt_even10k.biom`

josenavas:
to point to the corerct filepaths and to use the latest biom

Fixes #135 

I've just updated the notebook, is there anything else that I've to update? Also, are there tests for the functions on the ipython notebook?


wasade:
Thanks, @josenavas. The TYPES_OF_PLANTS plot is wrong, it looks like there is an issue with the metadata for a single sample somewhere. The quick fix for the notebook is to ignore sample sets in which the number of samples represented is 1. 

No tests, I've found it difficult to test inline methods in notebooks. Not opposed. Output looks correct with the exception of the TYPES_OF_PLANTS plot

wasade:
That field and surrounding metadata don't look obviously wrong. Issue is with sample 000001078.1075961. I suspect this was a manual update? Weird...

josenavas:
I just put a check in place to drop sample sets in which the number of samples represented is 1. I'm not sure what is going on with the metadata, but I hope that a space character is not the issue on that mapping file....

wasade:
it isn't. metadata look sane, I think this was manual edit but I'm not sure
how. I can't recall what is possible via the old admin interface

josenavas:
Also, as a note, I hat to include another check for ignoring empty metadata values (empty strings) as there was a few of those in all plots. So besides `no_data` and `NA` it also ignores empty strings. I do not know if that is a bug in another package or this is expected...

JWDebelius:
I usually use pass `na_values=['NA', 'no_data', 'unknown', '']`  to `pd.read_csv`.

wasade:
This is using the old `MetadataMap` that didn't take off. 

It isn't a bug, just that they need to be handled appropriately.

I think this is good to go unless there are any other issues?

josenavas:
Not that I'm aware of!

PR #137, opened 2015-03-24
Jsonized taxonomy results

wasade:
This produces JSON-based taxonomy summaries using nested objects. Each taxon is represented, as well as its count and relative frequency. The intent here is to provide a reasonably easily consumable format for web applications for representing per-sample taxonomy results. Newick is not ideal here as you have to hack it in order to embed metadata. JSON is not _great_ for trees due to the nesting, however taxonomy in bacteria and archaea top out at 7 levels (8 including root) and eukarya top out at I believe 14 (15 with root), so the nesting is not extreme.

Note, this is _not_ using scikit-bio's `TreeNode`. The taxtree code was put in place prior to scikit-bio, and we needed code that was BSD (hence not using cogent). It was easier to just adapt the existing code here than to do a larger refactoring. 

@madprime, just to keep you in the loop as it may be cool to provide these in the Open Humans pulldown. What I think we'll end up doing is hosting these via www.microbio.me as public downloads but I haven't resolved that specifically. What this then lets us do is provide relative abundance tab-delimited taxonomy tables and well as a computable format for the full taxonomy easily.

### DIFF
--- a/americangut/taxtree.py
+++ b/americangut/taxtree.py
@@ -161,6 +161,9 @@ def build_persample_tree_from_taxontable(table):
     abundance.
     """
     samp_ids = table.ids()
+
+    # It is assumed the IDs are of the form "foo; bar" or "foo;bar". Both of
+    # which have been produced by QIIME's summarize_taxa.py
     obs_ids_tmp = table.ids(axis='observation')
     obs_ids = [[t.strip() for t in i.split(';')] for i in obs_ids_tmp]
 
--- a/americangut/taxtree.py
+++ b/americangut/taxtree.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
-from biom.parse import parse_biom_table
+from itertools import izip
+
 
 __author__ = "Daniel McDonald"
 __copyright__ = "Copyright 2013, The American Gut Project"
@@ -11,9 +12,11 @@ __maintainer__ = "Daniel McDonald"
 __email__ = "mcdonadt@colorado.edu"
 
 
-def create_node(name):
+def create_node(name, **kwargs):
     """Create and return a new node"""
-    return {'name': name, 'popcount': 0, 'children': []}
+    n = {'name': name, 'children': []}
+    n.update(kwargs)
+    return n
 
 
 def add_node(cur_node, node):
@@ -38,7 +41,7 @@ def update_tree(tree, taxa_by_sample):
         children : list of nodes
     """
     if tree is None:
-        tree = create_node('root')
+        tree = create_node('root', popcount=0)
 
     for list_of_taxa in taxa_by_sample:
         tree['popcount'] += 1
@@ -53,7 +56,7 @@ def update_tree(tree, taxa_by_sample):
                 node = get_node(cur_node, taxon)
 
                 if node is None:
-                    node = create_node(taxon)
+                    node = create_node(taxon, popcount=0)
                     add_node(cur_node, node)
 
                 if taxon not in seen:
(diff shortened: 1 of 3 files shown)


americangut/taxtree.py
```diff
@@ -123,6 +153,28 @@ def build_tree_from_taxontable(table):
...
+def build_persample_tree_from_taxontable(table):
+    """Construct per-sample trees from a taxon table
+
+    yields (sample_id, tree). Each node in the tree (tip and nontip) include
+    the taxon name, associated sequence count as well as the relative
+    abundance.
+    """
+    samp_ids = table.ids()
+    obs_ids_tmp = table.ids(axis='observation')
+    obs_ids = [[t.strip() for t in i.split(';')] for i in obs_ids_tmp]
```
adamrp: Maybe make a note in the docstring that this is the expected format for the taxonomy metadata? biom supports multiple formats, iirc

scripts/ag
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
+
+import click
```
adamrp: `import click` should be down with `from biom import load_table`

americangut/taxtree.py
wasade: this is over the sample IDs, so there isn't a specific format. This consumes a table produced by qiime's summarize_taxa.py. Can add a mention if you'd like though?

adamrp: Woops, I misinterpreted this as taxonomy strings! A comment couldn't hurt, but not necessary

scripts/ag
```diff
@@ -0,0 +1,55 @@
...
+
+
+@ag.group()
+def taxtree():
+    pass
+
+
+@taxtree.command()
+@click.option('--table', required=True, type=click.Path(exists=True),
+              help='The input BIOM taxon table')
```
adamrp: I think you should also include `dir_okay=False` as an argument to `click.Path`

scripts/ag
```diff
@@ -0,0 +1,55 @@
...
+    for samp, ft, rare, uniq in it:
+        for r in rare:
+            click.echo("%s\t%s\t%s" % (samp, 'rare', r))
+        for u in uniq:
+            click.echo("%s\t%s\t%s" % (samp, 'unique', u))
+
+
+@taxtree.command()
+@click.option('--table', required=True, type=click.Path(exists=True),
+              help='The input BIOM taxon table')
```
adamrp: Same comment re: `dir_okay=False`

scripts/ag
```diff
@@ -0,0 +1,55 @@
...
+@click.option('--output', required=True, type=click.Path(exists=False))
```
adamrp: add `writable=True` to `click.Path` arguments

scripts/ag
```diff
@@ -0,0 +1,55 @@
...
+def persample_json(table, output):
+    os.makedirs(output)
```
adamrp: I didn't know this, but apparently [`makedirs`](https://docs.python.org/2/library/os.html#os.makedirs) gets confused if the argument contains [`os.pardir`](https://docs.python.org/2/library/os.html#os.pardir) (so, `..` probably). This seems like it could happen fairly regularly, so maybe call `os.path.abspath` on `output` before `os.makedirs`?

tests/test_taxtree.py
```diff
@@ -1,11 +1,15 @@
 #!/usr/bin/env python
 
 from unittest import TestCase, main
-from americangut.taxtree import create_node, add_node, get_node, update_tree, \
-        get_rare_unique, traverse, sample_rare_unique
-from biom.table import table_factory
+
 from numpy import array
 from copy import deepcopy
```
adamrp: `from copy import deepcopy` should be up with `from unittest import ...`

scripts/ag
wasade: that is surprising, good catch

adamrp:
Just a few comments

americangut/taxtree.py
wasade: done

scripts/ag
wasade: done

scripts/ag
wasade: done

scripts/ag
wasade: done

scripts/ag
wasade: done

scripts/ag
wasade: done

tests/test_taxtree.py
wasade: done

adamrp:
Thanks, @wasade! Do we not have Travis CI for this repo? I'm going to run the tests manually, but it would be nice to have Travis, too

wasade:
I don't think we do actually... this repo predates skbio if i remember, so
it is potentially still a bit lagging

adamrp:
I'll add an issue about it

adamrp:
Tests pass on my local system, merging

PR #139, opened 2015-03-25
Add .travis.yml

### LINKED ISSUE #138: Add Travis CI
We now have some test code, so it would be good to have Travis, too

adamrp:
Fix #138


wasade:
I'm nervous.

ElDeveloper:
I would suggest installing scipy, numpy, matplotlib and all the things that take a while with conda.

adamrp:
Okay, thanks for the tip @ElDeveloper! This is my first time adding travis to a repo, so I'm still figuring things out.

adamrp:
@wasade can you make your code (taxtree and test_taxtree) compatible with biom 1.3.1? If so, we are close to being able to add Travis (flake8 failures notwithstanding).

In a future PR, we should upgrade the package to use biom 2.x, but for now there is far more code that depends on the 1.3.1 API.

wasade:
The processing pipeline is QIIME 1.9, and much of the code in `americangut` is already BIOM 2.x. 

note, I _think_ `select_gamma` can be tossed as we've changed the filtering for blooms

adamrp:
There seem to be more errors when using BIOM 2.x than when I switch to BIOM
1.3.1, but I'll look into it.

adamrp:
Phew. Okay. It's passing. The vast majority of the changes are flake8 fixes (which i squashed into one commit, so for reviewing purposes it's probably safe to ignore the STY commit).

Note that right now, I have flake8 in Travis ignoring the py files in tests, since there are a *ton* of complaints...

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+before_install:
+  - wget http://repo.continuum.io/miniconda/Miniconda3-3.7.3-Linux-x86_64.sh -O miniconda.sh
+  - chmod +x miniconda.sh
+  - ./miniconda.sh -b
+  - export PATH=/home/travis/miniconda3/bin:$PATH
+  # Update conda itself
+  - conda update --yes conda
+install:
+  - conda create --yes -n env_name python=$PYTHON_VERSION pip nose numpy scipy matplotlib pandas>=0.15 flake8
+  - conda install --yes -n env_name python=$PYTHON_VERSION h5py>=2.3.1 statsmodels lxml
```
ElDeveloper: Any reason why these should be separated in two lines?

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - source activate env_name
+  # upgrade click, older versions of click do not have __version__
+  - pip install -U click
```
ElDeveloper: Same as above, all these can be put together.

americangut/make_phyla_plots.py
```diff
@@ -14,6 +13,9 @@
 from operator import itemgetter
 import colorbrewer
 
+
+use('agg')
```
ElDeveloper: You could leave this where it was (assuming moving this line may cause trouble later) by adding `#  noqa` at the end of that line. This tells flake8 to ignore any errors raised by this line.

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+  - pip install scikit-bio
+  - pip install biom-format
+  - pip install colorbrewer
+script:
+  - nosetests --with-doctest
+  - flake8 americangut/*.py
```
ElDeveloper: This and the next line could be put together.

ElDeveloper:
Looks good, a few minor comments.

.travis.yml
adamrp: one is `conda create` and the other is `conda install`; I don't think `create` will work if I put statsmodels and lxml on that line, will it?

americangut/make_phyla_plots.py
adamrp: It was complaining about the other lines (imports not being at top of file); will the `noqa` directive fix those complaints? Also, does the placement of this call matter with respect to the imports?

.travis.yml
adamrp: I actually found it helpful during debugging to have these separated, so that all of the flake8 failures pertaining to the scripts were lumped together and all the failures related to the library code was lumped together. But if they *should* be put together for some other reason, I can do that.

americangut/make_phyla_plots.py
ElDeveloper: Matplotlib recommends that this be set before any package makes an
import from pyplot: http://matplotlib.org/faq/usage_faq.html#what-is-a-backend

However I'm not sure if this is a problem with this package, thus I
suggested the noqa thing just to avoid the possibility of it being a
problem.

.travis.yml
ElDeveloper: Ah ok, I don't think they *have* to be together, but it seems strange
that they aren't. It's up to you, and there's no "travis.yml" guidelines
so yeah ...

.travis.yml
ElDeveloper: I believe it will work, that's what `.travis.yml` in qiita does.

adamrp: Ok I'll make that change, thanks

americangut/make_phyla_plots.py
adamrp: Ok, I'll give it a shot and see what Travis says. Fwiw, my local flake8
wasn't complaining about the imports not being at the top of the file but
Travis was, must be different versions or configuration

ElDeveloper: Yeah, I found something similar to this earlier this week, I think conda
has flake8 2.3.0 and pip has 2.4.0.

adamrp:
Ready for another review or merge

PR #142, opened 2015-03-31
Gradient legend maker

wasade:
Pretty light, just creates legends like this. Transparency is enabled by default as well. 

firmicutes_abundance


ElDeveloper:
:bomb: rainbow!

PR #150, opened 2015-05-16
All new PCoAs and legends for bottom half of Module 2 report, includi…

cuttlefishh:
All new PCoAs and legends for bottom half of Module 2 report, including latex and matplotlib code. Double-check that pipeline-generated macros files still work (I had to change the dummy macros_gut.tex file but I think this was due to changes already in the repo).  Still need to tweak a few pieces of the latex and python script.  Also need to add the actual rendering of the PCoAs via the pipeline, making sure the right options and file paths are invoked.


ElDeveloper:
Minor, but we should change the script to use spaces instead of tabs for
indentation.

cuttlefishh:
Cool. I will include spaces (not tabs) when I push the next set of edits.

wasade:
The spaces/tabs thing is required, it plays hell with the codebase if its not consistent. There are a large number of pep-8 failures btw

scripts/mod2_pcoa.py
```diff
@@ -0,0 +1,276 @@
...
+				  'Malawi': '#0000FF',
+				  'Netherlands': '#FF0000',
+				  'New Zealand': '#FF0000',
+				  'Norway': '#FF0000',
+				  'Scotland': '#FF0000',
+				  'Spain': '#FF0000',
+				  'Switzerland': '#FF0000',
+				  'Thailand': '#FF0000',
+				  'United Arab Emirates': '#FF0000',
+				  'United Kingdom': '#FF66FF',
```
wasade: only UK stands out? Does this hardcoded value need to change for AG?

scripts/mod2_pcoa.py
```diff
@@ -0,0 +1,276 @@
...
+				color_index = numeric_UK.index.tolist().index(sample)
+			else:
+				color_index = numeric_Other.index.tolist().index(sample)
+		except ValueError:
+			color_index = None
+
+		if color_index is None:
+			_color = '0.5'
+		else:
+			if mf.loc[sample].COUNTRY == 'United Kingdom':
```
wasade: same comment here

wasade: ...and line 248

scripts/mod2_pcoa.py
cuttlefishh: In the new version it's marked Western and then country of participant is changed to Highlight.

wasade:
Able to address the flake8 issues?

cuttlefishh:
I might need some assistance on the pep-8/flake8 issues.  Maybe @ElDeveloper can help this morning.

ElDeveloper:
I can help you out!

cuttlefishh:
I'm working on it. Didn't know what flake8 was, but Zech told me, I pip installed that shiznit, and am working through it now. Cool way to clean up code!

cuttlefishh:
OK, the latest updates should fix all known issues:
* flake8 has no remaining issues
* Figure 1 (body site) Z-order is now intercalated by slicing each study-body-site into 50 slices and randomly setting Z-order (each plot will look slightly different!)
* Figure 2 (country) Z-order is now set with other Western countries on bottom, then Malawi and Venezuela, then participant's country on top
* Figures 1, 2, and 3 have edge color as darker shade of dot color, not black (looks better when zoomed out)

Thanks to @ElDeveloper and @amnona for assistance and suggestions. @JWDebelius suggested to use alpha diversity instead of firmicutes for Figure 3, but that may have to wait (couldn't find it in any mapping file).

wasade:
two flake8 issues still FYI

wasade:
This looks great though. Getting the processing notebook setup to support it now.

cuttlefishh:
I don't get any flake8 issues when I run it on my laptop. Is there some version of flake8 that github runs?

wasade:
Should be using the latest version as no version is specified. Can you try doing:

```bash
pip install -U flake8
```

...and then rerun?

cuttlefishh:
I ran that update, updated from 2.4.0 to 2.4.1, but still no issues identified.  Are you just running with default options?

josenavas:
Just a note, I've seen this behavior, when I can't reproduce the flake8 errors locally. I think it's because the flake8 version installed with conda is different from the one in pip, but this is just an idea, no tests....

wasade:
@cuttlefishh, just look at the travis output in the PR. It'll show the reason the build fails

cuttlefishh:
OK, below is the line (used twice) creating the error.  The error is "line break before binary operator".  If I put it on the same line, it's more than 80 characters.  Not sure where I can break it legally.  Other option is I'm sure I can make the line shorter using a logical "or" as in "Malawi or Venezuela", but I am still new to Python logical structures.  Any suggestions from the experts?


        if (mf.loc[sample]['COUNTRY'] != 'Malawi') & (mf.loc[sample]['COUNTRY']
                                                      != 'Venezuela'):

josenavas:
```
if (mf.loc[sample]['COUNTRY'] != 'Malawi') &
        (mf.loc[sample]['COUNTRY'] != 'Venezuela'):
```

cuttlefishh:
Thanks Jose, but that gives me an "invalid syntax" error.

cuttlefishh:
This does not give me an error, but still don't know how it will go on github.

        if (mf.loc[sample]['COUNTRY'] != 'Malawi') & (
                mf.loc[sample]['COUNTRY'] != 'Venezuela'):

josenavas:
That should make it. Note that I've updated my comment just after I send it, cause I saw the syntax error (I just wrapped everything in `()`, but your current solution shouldn't be an issue.

cuttlefishh:
Cool, thanks! Still learning :)

wasade:
Ah, damn. Merged this prematurely. When looking at integrating, there are two things missing. First is that it needs an output directory. Second, it needs the ability to specify a prefix for the output files. I'll issue a separate PR on that shortly though.

cuttlefishh:
Yeah, I discussed that with @eldeveloper but we decided you would figure it out, as we didn't know how it would fit in with the rest of the processing pipeline. Thanks!

wasade:
Please describe incomplete PRs as a work in progress and explicitly enumerate what needs to be done in these instances.

What version of skbio did you write this against? The methods used do not exist.

cuttlefishh:
scikit-bio==0.2.3

wasade:
How difficult would it be to specify what samples are printed when called? Issue is that this script is slow and cannot be parallelized in its current form (at least, specifying a reduced mapping file causes it to break). I'm currently projecting 16 hours to produce the results for figure 1.

wasade:
nm, ill adjust it

cuttlefishh:
Yeah, we realized that it wasn't necessary to generate reports for HMP and GG, but didn't institute any way to filter those out.  Sounds like you got it working.

wasade:
No, still can't get the code to actually run yet...

cuttlefishh:
If speed of Figure 1 is the slowest part, you can speed that up by changing "50" in lines 60 and 63 to "20".  That will slice each category into 20 slices instead of 50, so there are fewer sets to plot.

wasade:
If that decreases the runtime by 10x then it would be acceptable, what's
your take there?

cuttlefishh:
Let me test that quickly on my laptop (I'm assuming it will scale on pando/compy).  We can reduce the number of slices even lower, but as we approach 1 we start to get points from one group (e.g. HMP fecal) covering up those from another group (e.g. AG fecal).

cuttlefishh:
If we set it to 10 slices, it takes ~1.5 seconds per plot on my laptop. Compare that to ~3 seconds per plot for 50 slices.  It still looks OK but not quite as well intercalated, and there's more plot-to-plot variation.  That's a 2x gain, not 10x.  Is that enough to be worth it?

wasade:
Drops it to ~8 hours by present estimates. Figured out the juggling
necessary to specify samples though, issuing a PR momentarily

cuttlefishh:
The challenge with using a reduced mapping file, I think, is the script assumes the sample IDs are the same (and in the same order) in the mapping file and PC file.

cuttlefishh:
Great. In that case, I would prefer to keep it at 50 slices because it looks much better.  Or compromise and set it to 25.

rob-knight:
If we’re talking about a 1 day difference, it’s worth it to make them nicer as this is already so much later than we wanted…

wasade:
Just issued #152 which allows this to be done in parallel. Walltime is now managable

cuttlefishh:
Looks good.  One nice thing about passing sample IDs as a parameter is we can quickly test known samples as a sanity check.  Probably worth doing before we deploy for the whole AG.

wasade:
That is a definite benefit as well

wasade:
...hooking it up now though

PR #151, opened 2015-05-18
Re-adding taxonomy

wasade:
Tables were missing taxonomy


JWDebelius:
The taxonomy was on the tables I downloaded. :+1:

adamrp:
Verified that the old tables did not have taxonomy observation metadata, and the new tables do :+1:

PR #154, opened 2015-07-31
Dump sample - accession mappings, and update for r1-15

wasade:
@ElDeveloper @adamrp, possible for quick review? This method isn't in place and the r1-15 sample mappings aren't in the repo


scripts/ag
```diff
@@ -54,5 +54,30 @@ def persample_json(table, output):
             fp.write(json.dumps(tree, indent=2))
 
 
+@ag.command()
+@click.argument('accession-metadata', required=True, nargs=-1)
+def sample_to_accession(accession_metadata):
+    """Dump {accession: [sample]} and {sample: accession}"""
```
ElDeveloper: Why is it a list of a single sample? Why not just the sample itself?

scripts/ag
```diff
@@ -54,5 +54,30 @@ def persample_json(table, output):
...
+    accession_to_sample = {}
+    sample_to_accession = {}
+
+    for f in accession_metadata:
+        accession = os.path.basename(f).rsplit('.', 1)[0]
+        accession_to_sample[accession] = []
+        with open(f) as fp:
```
ElDeveloper: Worth opening in 'U' mode?

ElDeveloper:
:scream: Travis says all checks failed, but the code looks good at a first glance. I don't have a dev environment to test this in though.

scripts/ag
wasade: It's `{accession: [list, of, samples]}`

scripts/ag
wasade: eh, we control their production from within the code base

scripts/ag
ElDeveloper: Ahhh, would you mind clarifying that doctor? :briefcase:

scripts/ag
ElDeveloper: Got it.

scripts/ag
wasade: Isn't that one of the normal notations to see? Like `[int]` implies `list` of `int`. Can change if you'd like tho

ElDeveloper: Hmmm, I guess I wasn't familiar with that notation, if it's standard
then I should revise my ways, so yeah just go with whatever is standard
(which seems like the current approach).

wasade:
Passing now, so up to you!! :)

PR #155, opened 2015-09-23
Subsampled figure2

wasade:
Performs the subsampling necessary so PCoA 2 isn't overwhelmed with US samples


wasade:
that was a fail...

PR #156, opened 2015-09-23
Subsample mod2

wasade:
Do the subsampling for PCoA 2 to make the patterns more clear


scripts/mod2_pcoa.py
```diff
@@ -37,6 +50,12 @@ def body_site(coords, mapping_file):
                      index_col='#SampleID')
     mf = mf.loc[o.site_ids]
 
+    if samples is None:
+        samples = mf.index
+    else:
+        samples = set(samples.split(',')).intersection(set(o.site_ids))
```
squirrelo: Should we allow sample names to be passed in that aren't in the file? I'd add at least a warning so we know if something like that happens.

scripts/mod2_pcoa.py
```diff
@@ -99,7 +167,13 @@ def country(coords, mapping_file):
     # mapping file
     mf = pd.read_csv(mapping_file, '\t', converters=defaultdict(str),
                      index_col='#SampleID')
-    mf = mf.loc[o.site_ids]
+    # mf = mf.loc[o.site_ids]
+
+    if samples is None:
+        samples = dm.ids[:]
+    else:
+        samples = set(samples.split(',')).intersection(set(dm.ids))
```
squirrelo: same question

scripts/mod2_pcoa.py
wasade: it isn't unusual for the mapping file to be a superset

squirrelo: I thought mf is the mapping file and o is the coordinates file. This may be interpretation issue, but if this is reading from o, you could theoretically have a subset of samples from the mapping file in o, but pass in the full mapping file sample list?

wasade: oh, i see now. sorry. this is to support task farming it as its embarassingly parallel

squirrelo: Ah, cool

squirrelo:
:+1:

wasade:
@squirrelo @ElDeveloper can this be merged?

squirrelo:
Looks good to me, I'll let @ElDeveloper take a final look.

PR #160, opened 2015-10-06
WIP: Preprocessing revitalized

wasade:
First pass. This has at its core a different approach to processing. The central idea is to not tie to a compute resource which greatly reduces the amount of code necessary to manage compute. Instead, the model is that the notebooks will assume they have full access to hammer on the machine they're executing on. This way, we can rely on making direct system calls (i.e., `!print_qiime_config.py`) as well as QIIME's parallelism within a single host through just forking multiple processes. On a compute cluster, the expectation is that an individual would request an entire node for a reasonable amount of walltime to do the work.

- [x] setup `runipys` similar to [IAB](https://github.com/gregcaporaso/An-Introduction-To-Applied-Bioinformatics/blob/master/runipynbs.py)
- [x] tie into `.travis.yml`, and set `export AG_TESTING=True`
- [x] resolve filtering notebook 
- [x] resolve OTU picking notebook
- [x] prepare for meta analyses (merging tables, mappings, etc) 
- [x] resolve diversity analysis notebook
- [x] resolve taxonomy summaries notebook
- [x] resolve generating of participant results notebook
- [ ] resolve project wide summary


wasade:
...also, this encourages much smaller notebooks which is refreshing to say the least.

americangut/util.py
```diff
@@ -20,6 +22,87 @@
 __email__ = "mcdonadt@colorado.edu"
```
mortonjt: Do you still want to keep this email address?

americangut/util.py
```diff
@@ -178,15 +261,29 @@ def fetch_metadata_xml(accession):
...
+
+    description = sample.find('DESCRIPTION')
+    metadata['Description'] = description.text.strip('" ')
+
     return metadata
 
 
-def fetch_study(accession, metadata_path, fasta_path):
+def fetch_study(accession):
     """Fetch and dump a full study
```
mortonjt: It'll be preferable to have a Parameters section here - up to you.

americangut/util.py
```diff
@@ -201,9 +298,10 @@ def fetch_study(accession, metadata_path, fasta_path):
         md = fetch_metadata_xml(sample)
         all_md[qiimedb_samplename] = md
         all_cols.update(md)
-
+        break
         # write out fasta
         try:
+            raise ValueError()
```
mortonjt: Since this is raising a ValueError, is the following couple of lines really necessary?  It looks like they are being short-circuited

wasade: thanks for catching. ya, intentionally broken right now.

americangut/util.py
wasade: this is old code, so i agree that refreshes on it are needed but not the immediate focus.

americangut/util.py
wasade: meh

ipynb/filter_sequences_for_blooms.ipynb
```diff
@@ -0,0 +1,246 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Early in the American Gut Project, it was observed that some organisms bloomed likely as a result of increased shipping time and delay between when samples were collected and when they were put on ice (more detail can be found [here](http://americangut.org/?page_id=277)). The purpose of this notebook is to apply the filter developed in order to bioinformatically subtract these observed bloom sequences from fecal samples. It is important to apply this filter when combining data with the American Gut as to remove a potential study-effect bias as all fecal data in the American Gut has had this filter applied. The specific steps covered are:\n",
```
mortonjt: This is a bit nit-picky.

But the term "apply the filter" is a little ambiguous at first, since the "filter" has not been identified.

Is it possible to reword it to something like
"The purpose of this notebook is to illustrate how these organisms are identified and filtered out in our bioinformatics software pipeline."

ipynb/pick_otus.ipynb
```diff
@@ -0,0 +1,228 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Pick OTUs (for full and trimmed data) at approximately genus level resolution (97% similarity) using SortMeRNA, closed reference against Greengenes 13_8."
```
mortonjt: Just to sanity check.

Can we assume that people looking at these notebooks have prior knowledge about OTU picking?

If not, it may be desirable to at least include a link.

ipynb/pick_otus.ipynb
```diff
@@ -0,0 +1,228 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "filtered_sequences       = agu.get_existing_path(agenv.paths['filtered-sequences'])\n",
+    "filtered_sequences_100nt = agu.get_existing_path(agenv.paths['filtered-sequences-100nt'])\n",
```
mortonjt: Do we want to mention what are trimmed sequences and how the trimming was done?

ipynb/pick_otus.ipynb
```diff
@@ -0,0 +1,228 @@
...
+      "    main()\n",
+      "  File \"/Users/mcdonadt/miniconda3/envs/agdev/bin/pick_closed_reference_otus.py\", line 224, in main\n",
+      "    status_update_callback=status_update_callback)\n",
+      "  File \"/Users/mcdonadt/miniconda3/envs/agdev/lib/python2.7/site-packages/qiime/workflow/upstream.py\", line 506, in run_pick_closed_reference_otus\n",
+      "    close_logger_on_success=close_logger_on_success)\n",
+      "  File \"/Users/mcdonadt/miniconda3/envs/agdev/lib/python2.7/site-packages/qiime/workflow/util.py\", line 122, in call_commands_serially\n",
+      "    raise WorkflowError(msg)\n",
+      "qiime.workflow.util.WorkflowError: \n",
+      "\n",
+      "*** ERROR RAISED DURING STEP: Make OTU table\n",
```
mortonjt: Did this fail?  If so, will this need to be corrected?

mortonjt:
I think this is mostly okay.  There are a few comments in pick_otus.

americangut/util.py
```diff
@@ -20,6 +22,87 @@
...
+    """
+    path = get_path(path)
+
+    if not os.path.exists(path):
+        raise IOError('%s does not exist.' % path)
+
+    return path
+
+
 def pick_rarifaction_level(id_, lookups):
```
gregcaporaso: spelling error, should be *rarefaction*

gregcaporaso:
*Remove the recruited bloom sequences from the demultiplexed sequence data*

Why do this at the sequence level and not the OTU level? (And if this isn't the kind of feedback you're looking for now, just let me know.)

rob-knight:
Because doing it at the OTU level was too broad in earlier tests, though perhaps this should be revisited.

gregcaporaso:
*We also need to specify what specific metadata category and value correspond indicate what samples are fecal.* extra words here

gregcaporaso:
The bloom sequences are used as reference sequences for 97% closed-reference OTU picking, and then all sequences that cluster into the OTUs defined by those centroids are removed. So the "range" of sequences that are removed is the same size as if you did this at the OTU level, but the centroids are specifically defined by the bloom sequences. This is interesting and makes sense, but I think it'd be good to describe this specifically (where you mention how sequences are recruited above) as it wasn't the way that I would have thought to do this.

gregcaporaso:
One really high-level comment on the notebooks - if these were written in markdown instead of ipynb they'd be easier to diff, have others contribute to, etc. This is what we're doing with IAB now (see [my MSL blog post](https://www.mozillascience.org/an-introduction-to-applied-bioinformatics-at-mozsprint) about this).

wasade:
That is awesome. Yes, will look at that right now

gregcaporaso:
[This video](https://www.youtube.com/watch?v=s9-aZrX5CY8) shows how readers can submit changes as PRs 100% through their web browser.

gregcaporaso:
Reviewed ``get_sequences_and_metadata.ipynb``, that all makes sense.

gregcaporaso:
In ``pick_otus.ipynb``, you shouldn't need to pass ``-r`` or ``-t`` anymore - they're defaulting to the files you're wanting to use (I think).

gregcaporaso:
It could help with debugging if your two ``pick_closed_reference_otus.py`` commands were in different cells.

gregcaporaso:
@wasade, added some comments on your notebooks. These generally look good to me (clearly *a lot* better than the massive notebook you linked me to). If you want me to review a few more of these as you get them together let me know.

wasade:
Not fully, I want to be able to exercise these notebooks via travis so I
need to replace the reference used as GG 97% isn't feasible for testing due
to 15min indexing time with SortMeRNA (and not able to incl. in repo due to
size)

wasade:
Thank you @gregcaporaso, this is awesome. I'm going to finish doing the indirection to get the testing up for pick OTUs, tie into .travis, and then do the ipymd bit you linked.

gregcaporaso:
Ah, right, that makes sense (though you could probably add a .qiime_config
to travis if you wanted to, but no difference either way).

americangut/notebook_environment.py
```diff
@@ -0,0 +1,135 @@
...
+    'gg-otus-100nt': 'otus/gg-13_8-97-percent-otus-with-100nt',
+    'gg-otus-biom': 'otus/gg-13_8-97-percent-otus/otu_table.biom',
+    'gg-otus-100nt-biom':
+        'otus/gg-13_8-97-percent-otus-with-100nt/otu_table.biom',
+}
+
+
+def _assert_environment():
+    if qiime.__version__ != '1.9.1':
+        raise ImportError("QIIME 1.9.1 is not in the environment.")
```
squirrelo: Maybe put current version info in the error?

wasade: :+1:

ipynb/primary-processing/3-pick_otus.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+Before we go too far, let's make sure the files we need are present.
+
+```python
+>>> filtered_sequences       = agu.get_existing_path(agenv.paths['filtered-sequences'])
+>>> filtered_sequences_100nt = agu.get_existing_path(agenv.paths['filtered-sequences-100nt'])
+```
+
+And, let's make sure that the output files we need do not already exist.
```
squirrelo: output files we create?

squirrelo:
Couple small things.

ipynb/index.ipynb
```diff
@@ -4,12 +4,20 @@
...
     "\n",
-    "* Primary processing pipeline [Notebook](module2_v1.0.ipynb)\n",
-    "* Filter bloom sequences [Notebook](FilterAndPickOTUs.ipynb)\n",
+    "1. [Get sequences and metadata](primary-processing/1-get_sequences_and_metadata.ipynb)\n",
+    "2. [Filter sequences for blooms](primary-processing/2-filter_sequences_for_blooms.ipynb)\n",
+    "3. [Pick OTUs](primary-processing/3-pick_otus.ipynb)\n",
+    "4. [Prepare metaanalyses](primary-processing/4-prepare_metaanalyses.ipynb)\n",
+    "5. [Alpha diversity calculations](primary-processing/5-alpha_diversity.ipynb)\n",
+    "6. [Beta diversity calculations](primary-processing/6-beta_diversity.ipynb)\n",
+    "7. [Taxonomy summaries](primary-processing/7-taxonomy_summaries.ipynb)\n",
```
wasade: needs 7.5

americangut/notebook_environment.py
```diff
@@ -0,0 +1,355 @@
...
+
+    if find_executable('print_qiime_config.py') is None:
+        raise EnvironmentError("The QIIME executables are not in $PATH.")
+
+    if find_executable('mod2_pcoa.py') is None:
+        raise EnvironmentError("The AG scripts are not in $PATH.")
+_assert_environment()
```
squirrelo: Is this supposed to be here?

wasade: yes. if the environment is not sane, the bail as early as possible

americangut/parallel.py
```diff
@@ -0,0 +1,31 @@
...
+        A file-like object to write a list of successful sample IDs too
+    fail_fp : file-like object
+        A file-like object to write a list of unsuccessful sample IDs too,
+        and any associated error information
+    partitions : iterable
+        Yields a function and an iterable of IDs. It is expected that this
+        function will return a dict of {str: list}.
+    """
+    pool = mp.Pool(processes=mp.cpu_count())
+
```
mortonjt: Would it be worthwhile to pass in the number of cores as a parameter?

That why the user can specify exactly how many cores that they would like to use.  This could also help prevent users from accidentally over-swamping their system.

wasade: It introduces either a config file or yet another env var.

wasade:
In the interest of simplifying review, as this PR is titanic, I'm going to close it out and open individual PRs for each notebook one-by-one.

Issue #161, opened 2015-10-21
Primary processing revitalized


wasade:
- [x] setup `runipys` similar to [IAB](https://github.com/gregcaporaso/An-Introduction-To-Applied-Bioinformatics/blob/master/runipynbs.py)
- [x] tie into `.travis.yml`, and set `export AG_TESTING=True`
- [x] get sequence and metadata (#162)
- [x] resolve filtering notebook (#163)
- [x] resolve OTU picking notebook (#164)
- [x] prepare for meta analyses (#165) 
- [x] resolve alpha diversity analysis notebook (#166)
- [x] resolve beta diversity analysis notebook (#167)
- [x] resolve taxonomy summaries notebook (#168)
- [x] resolve categories collapse (#169)
- [x] resolve generating of participant results notebook (#170)
- [ ] resolve project wide summary

jwdebelius:
Project wide summary is redundant with #61

PR #162, opened 2015-10-21
1 get sequences and metadata

wasade:
Addresses the first few points in #161


wasade:
This first PR for #161 is large as it brings over the infrastructure that was necessary for the other notebooks that had been developed. It would have been really painful to slice up the python modules. But, once this is merged, the other notebooks will have much less burden associated.

wasade:
@ElDeveloper @josenavas @squirrelo @mortonjt or @antgonza, possible for a review?

americangut/util.py
```diff
@@ -20,21 +23,85 @@
...
+
+def get_new_path(path):
+    """Get a new relative path to the working directory
+
+    Parameters
+    ----------
+    path : str
+        A path that does not exist
+
+    Notes
```
antgonza: The Notes should be after the Returns, right?

americangut/util.py
```diff
@@ -20,21 +23,85 @@
...
+    ----------
+    path : str
+        A path that exists
 
-    None is returned if the key is not found
+    Notes
+    -----
+    It is only assured that the path exists at the time of function evaluation
+
+    Raises
```
antgonza: same with Raises, no?

americangut/notebook_environment.py
```diff
@@ -0,0 +1,357 @@
...
+    -----
+    If $AG_TESTING == 'True', then the data returned will correspond to the
+    test dataset.
+
+    Raises
+    ------
+    IOError
+        If the filepaths are not accessible
+
+    Returns
```
squirrelo: This should probably get moved to right after parameters.

americangut/notebook_environment.py
```diff
@@ -0,0 +1,357 @@
...
+
+def get_accessions():
+    """Get the accessions to use, or redirect to test data
+
+    Notes
+    -----
+    If $AG_TESTING == 'True', then the accessions returned will
+    correspond to the test dataset.
+
+    Returns
```
squirrelo: Returns should be before notes.

americangut/notebook_environment.py
```diff
@@ -0,0 +1,357 @@
...
+
+def get_bloom_sequences():
+    """Get the filepath to the bloom sequences
+
+    Raises
+    ------
+    IOError
+        If the path does not exist
+
+    Returns
```
squirrelo: same here

americangut/util.py
```diff
@@ -20,21 +23,85 @@
...
+    Parameters
+    ----------
+    path : str
+        A path
+
+    Notes
+    -----
+    This method does not care if the path exists or not
+
+    Returns
```
squirrelo: same here

americangut/util.py
```diff
@@ -178,15 +245,29 @@ def fetch_metadata_xml(accession):
     metadata = {}
     for node in attributes.iterfind('SAMPLE_ATTRIBUTE'):
         tag, value = node.getchildren()
-        metadata[tag.text.strip('" ').upper()] = value.text.strip('" ')
+        if value.text is None:
+            metadata[tag.text.strip('" ').upper()] = 'no_data'
```
squirrelo: Any need in these to strip single quote as well, since already stripping?

americangut/util.py
```diff
@@ -285,3 +366,109 @@ def count_samples(metadata_fp, criteria=None):
...
+        sample_id = new_line[0]
+        body_site = new_line[bodysite_idx]
+        country = new_line[country_idx]
+
+        # grab the body site
+        if body_site.startswith('UBERON_'):
+            body_site = body_site.split('_', 1)[-1].replace("_", " ")
+        elif body_site.startswith('UBERON:'):
+            body_site = body_site.split(':', 1)[-1]
+        elif body_site in ['NA', 'unknown', '', 'no_data']:
```
squirrelo: use a set instead of list here for comparison?

squirrelo:
few comments and questions.

americangut/util.py
mortonjt: It shouldn't matter?  Since there are only 4 elements

.travis.yml
```diff
@@ -2,22 +2,30 @@ language: python
 env:
   - PYTHON_VERSION=2.7
 before_install:
-  - wget http://repo.continuum.io/miniconda/Miniconda3-3.7.3-Linux-x86_64.sh -O miniconda.sh
+  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
```
ElDeveloper: :clap:

americangut/util.py
```diff
@@ -285,3 +366,109 @@ def count_samples(metadata_fp, criteria=None):
...
+
+def clean_and_reformat_mapping(in_fp, out_fp, body_site_column_name,
+                               exp_acronym):
+    """Simplify the mapping file for use in figures
+
+    in_fp : input file-like object
+    out_fp : output file-like object
+    body_site_column_name : specify the column name for body
+    exp_acronym : short name for the study
+
```
mortonjt: I would personally prefer to use numpy-docs, your call.

americangut/__init__.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+working_dir = 'agp_processing'
```
ElDeveloper: Shouldn't this be all-caps?

americangut/notebook_environment.py
```diff
@@ -0,0 +1,357 @@
...
+    'successful-ids': '8/successful_ids.txt',
+    'unsuccessful-ids': '8/unsuccessful_ids.txt',
+    'per-sample-results': '8/per-sample-results',
+}
+
+
+def _assert_environment():
+    if qiime.__version__ != '1.9.1':
+        obs_version = qiime.__version__
+        raise ImportError(("QIIME 1.9.1 is not in the environment, found "
```
ElDeveloper: There's no need for the double parenthesis.

americangut/util.py
squirrelo:
```python
In [1]: x = ['string1', 'string2', 'string3', 'string4']

In [2]: y = {'string1', 'string2', 'string3', 'string4'}

In [3]: timeit lambda r: r in x
The slowest run took 20.53 times longer than the fastest. This could mean that an intermediate result is being cached 
10000000 loops, best of 3: 58.1 ns per loop

In [4]: timeit lambda r: r in y
The slowest run took 32.97 times longer than the fastest. This could mean that an intermediate result is being cached 
10000000 loops, best of 3: 57.9 ns per loop
```
Probably not, but is good for consistency.

americangut/notebook_environment.py
```diff
@@ -0,0 +1,357 @@
...
+    not already exist.
+    """
+    path = os.path.join(ag.working_dir, chp)
+    if not os.path.exists(path):
+        os.mkdir(path)
+    return path
+
+
+def get_sortmerna_index():
+    """Return the absolute path a SortMeRNA index if available"""
```
ElDeveloper: `a` -> `of a`

americangut/util.py
```diff
@@ -285,3 +366,109 @@ def count_samples(metadata_fp, criteria=None):
...
+        # grab the body site
+        if body_site.startswith('UBERON_'):
+            body_site = body_site.split('_', 1)[-1].replace("_", " ")
+        elif body_site.startswith('UBERON:'):
+            body_site = body_site.split(':', 1)[-1]
+        elif body_site in ['NA', 'unknown', '', 'no_data']:
+            errors[('unspecified_bodysite', body_site)].append(sample_id)
+            continue
+        else:
```
antgonza: Couldn't find a test for this raise.

americangut/notebook_environment.py
```diff
@@ -0,0 +1,357 @@
...
+        The path to the activated directory
+
+    Notes
+    -----
+    Activation creates the chapter processing directory if it does
+    not already exist.
+    """
+    path = os.path.join(ag.working_dir, chp)
+    if not os.path.exists(path):
+        os.mkdir(path)
```
ElDeveloper: Maybe worth using skbio's create_dir instead? It won't fail if the directory already exists.

```
>>> from skbio.util import create_dir
>>> create_dir('plelercosaster')
0
>>> create_dir('plelercosaster')
1
```

americangut/__init__.py
wasade: Maybe? No strong feelings..

americangut/notebook_environment.py
wasade: excellent point

antgonza: Just to be sure that are there? :smile:

americangut/notebook_environment.py
```diff
@@ -0,0 +1,357 @@
...
+    else:
+        return "1000"
+
+
+def get_reference_set():
+    """Get the reference set to use for OTU picking
+
+    Returns
+    -------
+    (str, str)
```
ElDeveloper: I don't think this is the way numpy-docs are supposed to be formatted, IIRC descriptions should be individual per return  value.

americangut/__init__.py
wasade: done

americangut/notebook_environment.py
wasade: done

americangut/notebook_environment.py
wasade: done

americangut/notebook_environment.py
wasade: done

americangut/notebook_environment.py
wasade: I disagree. Is that part of the numpydoc spec?

americangut/util.py
```diff
@@ -214,7 +295,7 @@ def fetch_study(accession, metadata_path, fasta_path):
     md_f.write('\t'.join(header))
     md_f.write('\n')
     for sampleid, values in all_md.iteritems():
-        to_write = [values.get(k, "no_data") for k in header]
+        to_write = [values.get(k, "no_data").encode('utf-8') for k in header]
```
ElDeveloper: Would you mind elaborating on why this encode is needed?

americangut/notebook_environment.py
wasade: guess np is doing that. odd.

americangut/notebook_environment.py
wasade: done

americangut/util.py
wasade: done

americangut/util.py
wasade: done

americangut/notebook_environment.py
squirrelo: There's an order of information given in [the howto](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections) but not actually explicit in whether that order must be maintained.

americangut/util.py
wasade: done

americangut/util.py
wasade: Haven't seen a need

americangut/util.py
wasade: This is old code/doc so not modifying unless necessary

americangut/util.py
wasade: consistency with what? the cost of a linear search on a small list is really small and can easily begin to outweight the cost of computing a hash. but this isn't performance bound code.

americangut/util.py
wasade: This is existing code, not changing unless necessary

americangut/notebook_environment.py
wasade: Doesn't appear to be in 0.2.3?

```python
12:09:09 (mcdonadt@8086):primary-processing> from skbio import create_dir
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-b7413eac0347> in <module>()
----> 1 from skbio import create_dir

ImportError: cannot import name create_dir
```

wasade: perhaps I should use the correct import...

wasade: while it would work, it just reduces 2 LOC to 1 and it confounds the behavior without knowing details of `create_dir`

americangut/notebook_environment.py
wasade: done

americangut/util.py
wasade: likely due to some utf-8 chars slipping into EBI at some point during AG's lifecycle

americangut/notebook_environment.py
wasade: Since system calls are made to QIIME, and are made somewhat late in
execution, the intent is to fail early and immediately if the environment
does not appear to be sane.

On Wed, Oct 21, 2015 at 12:00 PM, Antonio Gonzalez <notifications@github.com

americangut/util.py
squirrelo: Fair enough, especially if not performance bound.

americangut/notebook_environment.py
ElDeveloper: > Just to be sure that are there? :smile:

wasade:
Missed one instance of `working_dir`, good to merge following tests passing?

I promise the next set of notebooks will be much lighter, at least until we get to the per-sample stuff in notebook 8

ElDeveloper:
:+1:

wasade:
merge?

wasade:
thanks!

PR #163, opened 2015-10-21
filter sequences for blooms

wasade:
@ElDeveloper @antgonza @mortonjt @josenavas @squirrelo, round 2 is up.

### DIFF
--- a/ipynb/primary-processing/2-filter_sequences_for_blooms.md
+++ b/ipynb/primary-processing/2-filter_sequences_for_blooms.md
@@ -1,0 +1,102 @@
+Early in the American Gut Project, it was observed that some organisms bloomed likely as a result of increased shipping time and delay between when samples were collected and when they were put on ice (more detail can be found [here](http://americangut.org/?page_id=277)). The purpose of this notebook is to apply the filter developed in order to bioinformatically subtract these observed bloom sequences from fecal samples. It is important to apply this filter when combining data with the American Gut as to remove a potential study-effect bias as all fecal data in the American Gut has had this filter applied. The specific steps covered are:
+
+* Filter demultiplexed sequence data to only fecal samples
+* Determine what sequences in the fecal samples recruit to the observed bloom sequences
+* Remove the recruited bloom sequences from the demultiplexed sequence data
+* Trim the filtered demultiplexed data (to reduce study bias when combining with short reads, such as the data in [Yatsunenko et al 2012](http://www.nature.com/nature/journal/v486/n7402/abs/nature11053.html))
+
+
+The filtering is only intended to be applied to fecal data. As such, this notebook allows you to describe what metadata column and value to use so that only fecal samples are used.
+
+```python
+>>> import os
+>>> import multiprocessing
+>>> import americangut.notebook_environment as agenv
+>>> import americangut.util as agu
+...
+>>> chp_path = agenv.activate('2')
+```
+
+Next, we'll establish the paths we will be creating.
+
+```python
+>>> fecal_sequences          = agu.get_new_path(agenv.paths['fecal-sequences'])
+>>> filtered_sequences       = agu.get_new_path(agenv.paths['filtered-sequences'])
+>>> filtered_sequences_100nt = agu.get_new_path(agenv.paths['filtered-sequences-100nt'])
+>>> observed_blooms          = agu.get_new_path(agenv.paths['observed-blooms'])
+>>> observed_blooms_biom     = agu.get_new_path(agenv.paths['observed-blooms-biom'])
+>>> observed_blooms_otu_map  = agu.get_new_path(agenv.paths['observed-blooms-otu-map'])
+```
+
+This next call will setup and verify the path to the bloom sequences used for filtering.
+
+```python
+>>> bloom_sequences = agenv.get_bloom_sequences()
+```
+
+Now let's setup the paths to the sequences to filter. We need the metadata as well in order to reduce the data to just the fecal samples. Please replace these variables with your own paths if you wish to filter your data for blooms (a necessary precursor if you wish to combine data with the American Gut).
+
+```python
+>>> # If you are filtering your own data, please update these filepath variables as necessary
+... sequences = agenv.get_existing_path(agenv.paths['raw-sequences'])
+>>> metadata  = agenv.get_existing_path(agenv.paths['raw-metadata'])
+```
+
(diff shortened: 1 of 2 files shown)


ipynb/primary-processing/2-filter_sequences_for_blooms.md
```diff
@@ -0,0 +1,102 @@
+Early in the American Gut Project, it was observed that some organisms bloomed likely as a result of increased shipping time and delay between when samples were collected and when they were put on ice (more detail can be found [here](http://americangut.org/?page_id=277)). The purpose of this notebook is to apply the filter developed in order to bioinformatically subtract these observed bloom sequences from fecal samples. It is important to apply this filter when combining data with the American Gut as to remove a potential study-effect bias as all fecal data in the American Gut has had this filter applied. The specific steps covered are:
+
+* Filter demultiplexed sequence data to only fecal samples
+* Determine what sequences in the fecal samples recruit to the observed bloom sequences
+* Remove the recruited bloom sequences from the demultiplexed sequence data
+* Trim the filtered demultiplexed data (to reduce study bias when combining with short reads, such as the data in [Yatsunenko et al 2012](http://www.nature.com/nature/journal/v486/n7402/abs/nature11053.html))
```
antgonza: Should this be in a separate md? Note that this is not part of the bloom filtering ...

wasade: Good catch, copy pasta. Trimming is done in the pick OTUs step

ipynb/primary-processing/2-filter_sequences_for_blooms.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+The filtering is only intended to be applied to fecal data. As such, this notebook allows you to describe what metadata column and value to use so that only fecal samples are used.
```
antgonza: Where do you define that? I guess that should be just after this line.

ipynb/primary-processing/2-filter_sequences_for_blooms.md
antgonza: I want pasta, it's almost lunch time ...

ipynb/primary-processing/2-filter_sequences_for_blooms.md
```diff
@@ -0,0 +1,102 @@
...
+...                  --valid_states $_fecal_states
+```
+
+The next thing we need to do is setup the parameters for SortMeRNA, which is the method we'll use to compare all the input data to our reference of bloom sequences.
+
+```python
+>>> _params_file = agu.get_path('sortmerna_pick_params.txt')
+>>> with open(_params_file, 'w') as f:
+...     f.write("pick_otus:otu_picking_method sortmerna\n")
+...     f.write("pick_otus:similarity 0.97\n")
```
antgonza: Isn't this the default?

ipynb/primary-processing/2-filter_sequences_for_blooms.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+As the data have now been filtered for blooms, we can now trim the reads back to 100nt to minimize a potential study effect when combining with the Global Gut.
+
+```python
+>>> with open(filtered_sequences) as in_, open(filtered_sequences_100nt, 'w') as out:
+...     agu.trim_fasta(in_, out, 100)
+```
+
+Finally, let's do a quick sanity check that we have sequence data and we'll also dump out summary information about how many reads per sample recruited to the blooms.
```
antgonza: I get that this is a sanity check but > than zero doesn't say much. What about testing same number of seqs or something like that ...

antgonza:
A few comments ... nothing major.

ipynb/primary-processing/2-filter_sequences_for_blooms.md
wasade: What would the expected number be?

wasade:
Thanks!

ipynb/primary-processing/2-filter_sequences_for_blooms.md
antgonza: As many as agu.trim_fasta writes?

ipynb/primary-processing/2-filter_sequences_for_blooms.md
wasade: this is actually done in this notebook at the end. i dont see the benefit in a notebook dedicated to trimming though...

antgonza: Yes, saw that. Perhaps adding to the main description. Note that it talks about the bloom filtering but not about the length.

ipynb/primary-processing/2-filter_sequences_for_blooms.md
wasade: What I don't like about this is the extensive amount of time it'll take to walk over the 2 x 30GBs of sequence data.

wasade: Up to you, can make the change if you'd like

antgonza: What about removing the > than zero tests?

wasade: So not performing any sanity checking?

antgonza: I guess the biom summary could suffice. Again, not strong feelings about this but IMOO a sanity check of file > than 0 is not really sanity, right?

wasade: The `observed_blooms_biom` file is not used anywhere after this notebook. Checking that a file is nonzero tells you it has data, which is what we expect. We do not have expectations on the number of sequence counts as that will change over time, although we could assert that there are at least 10000 sequences but that also doesn't really tell you much. Testing if the number of reads is the same between these files tells you a little more than the nonzero test but all it says is that the data appear to be fasta-like. However, we already expect that as we assume QIIME's `filter_fasta.py` script works and similarly, we expect `trim_fasta` to work as it is both extremely simple and covered by an exercised unit test. 

If the files that result from this notebook are empty, it'll cause confusion on the pick OTUs phase, so the goal is to identify possible issues as early as is feasible.

wasade: And actually, testing if the number of reads is the same would still pass if both files were empty

mortonjt:
I don't see any immediate problems.  Is this ready to merge?

wasade:
If @antgonza gives a :+1:

wasade:
thanks!

PR #171, opened 2015-10-24
Adding in setup.py

mortonjt:


setup.py
```diff
@@ -0,0 +1,58 @@
...
+    Programming Language :: Python :: 3
+    Operating System :: Unix
+    Operating System :: POSIX
+    Operating System :: MacOS :: MacOS X
+"""
+classifiers = [s.strip() for s in classes.split('\n') if s]
+
+setup(name='americangut',
+      version=version,
+      description="American gut analyses",
```
wasade: gut -> Gut

setup.py
```diff
@@ -0,0 +1,58 @@
...
+      long_description="American Gut open-access data "
+                       "and IPython notebooks",
```
wasade: notebooks -> Notebooks

setup.py
```diff
@@ -0,0 +1,58 @@
...
+      author="American Gut development team",
+      author_email="jamietmorton@gmail.com",
```
wasade: it would be more appropriate to list either me, Justine, Embriette or Rob

setup.py
```diff
@@ -0,0 +1,58 @@
...
+      maintainer="American Gut development team",
+      maintainer_email="jamietmorton@gmail.com",
```
wasade: Same here unless you'd like to be the maintainer :)

setup.py
```diff
@@ -0,0 +1,58 @@
...
+      url='http://github.com/biocore/American-Gut/',
```
wasade: http -> https

wasade:
Few comments, :+1: otherwise once tests pass

setup.py
wasade: ...actually, Just use "American Gut development team" and the American Gut help account email.

mortonjt:
:boom: tests passing

squirrelo:
Already merged now, but did we really need a setup.py given that this is primarily a collection of notebooks and data? Would a pip requirements.txt and/or the conda equivalent have been more appropriate?

wasade:
It makes deployment easier

squirrelo:
eh, 
```bash
pip install .
```
vs 
```bash
pip install -r requirements.txt
echo $(pwd)/scripts >  .bash_profile
```
The second has an extra line, but also doesn't treat the repo as a full installable package. I guess the question for me is, should this be treated as a full installable package or a collection of notebooks?

wasade:
What's the downside to it being installable?

PR #173, opened 2015-10-24
WIP: 10 participant results

wasade:
Get back to generating participant PDFs. 

Note: the figure generating scripts now operate only on single samples at a time. Ideally these transition to in memory operations as opposed to system calls, but those aspects are abstracted out right now so the only concern is system call overhead which is large but not prohibitive.


wasade:
Participant results are passing! 

Do not merge yet though, still doing validation locally and on barnacle

americangut/per_sample.py
```diff
@@ -228,6 +230,40 @@ def taxa_summaries(opts, sample_ids):
     return results
 
 
+def check_sequence_counts(opts, sample_ids):
+    """Errors if the sequence counts post filtering are < 1000
```
squirrelo: This description seems inaccurate. This gives count info if under rarefaction_depth, correct, not errors?

scripts/mod2_pcoa.py
```diff
@@ -299,55 +294,50 @@ def gradient(coords, mapping_file, color, output, prefix, samples):
...
-        out_file = os.path.join(output, '.'.join([prefix, sample, 'pdf']))
-        plt.savefig(out_file, figsize=figsize, dpi=my_dpi)
-        plt.close()
+    # plot numeric metadata as colored gradient
+    ids = numeric.index
+    x, y = c_df.loc[ids][0], c_df.loc[ids][1]
+    plt.scatter(x, y, c=numeric[color], cmap=plt.get_cmap('RdBu'),
+                alpha=ALPHA, lw=LINE_WIDTH, edgecolor=color_array*0.6)
+
+    # plt.colorbar()
```
squirrelo: is this needed or can it be removed?

squirrelo:
Two small comments

PR #176, opened 2015-10-28
Fixes the double SampleID Column

JWDebelius:
There are duplicate `#SampleID` column in the rounds 1-21 data.
This prevents the duplicate column by splitting the barcode into a unique column, and the using that to key.


wasade:
:+1: pending tests

mortonjt:
Thanks for catching/fixing this @JWDebelius !  Will merge once the tests pass

PR #177, opened 2015-11-06
WIP: Do the 10k

wasade:
This is a work in progress that refactors path management, and includes support for generating many other datasets of interest for downstream use.

Path management is not complete yet as, in the ideal case, there is a data access object to facilitate developers (e.g., `AGData.bdiv('notrim', '1k', 'fecal')` to get back the corresponding `DistanceMatrix`). In addition, the shear number of paths tracked is large but the patterns suggest it should be possible to not explicitly define them. However, right now, development is on a near path of least resistance.

PR #179, opened 2015-11-10
WIP: Create Alpha Diversity violin plot code

JWDebelius:
This is the code to create alpha diversity violin plots.

I've updated the alpha diversity notebook to specify that we also want shannon diversity calculated.
There is code to do body-site specific plotting. It needs to be moved over to a click command, and somehow incorporated in the primary processing notebook.


wasade:
It isn't passing yet, but could you merge in #177? It does shannon and 10k, etc

PR #184, opened 2015-12-15
add Unknown to unspecified body site list

EmbrietteH:
This enables primary processing of the notebooks, preventing failure on notebook 04 (metadata) due to blanks having entries as "Unknown." With the latest EBI update, the values are now first-letter capitalized and the code needs to be updated accordingly.

### DIFF
--- a/americangut/util.py
+++ b/americangut/util.py
@@ -492,7 +492,7 @@ def clean_and_reformat_mapping(in_fp, out_fp, body_site_column_name,
             body_site = body_site.split('_', 1)[-1].replace("_", " ")
         elif body_site.startswith('UBERON:'):
             body_site = body_site.split(':', 1)[-1]
-        elif body_site in ['NA', 'unknown', '', 'no_data', 'None']:
+        elif body_site in ['NA', 'unknown', '', 'no_data', 'None', 'Unknown']:
             errors[('unspecified_bodysite', body_site)].append(sample_id)
             continue
         else:


squirrelo:
:+1:

americangut/util.py
```diff
@@ -492,7 +492,7 @@ def clean_and_reformat_mapping(in_fp, out_fp, body_site_column_name,
             body_site = body_site.split('_', 1)[-1].replace("_", " ")
         elif body_site.startswith('UBERON:'):
             body_site = body_site.split(':', 1)[-1]
-        elif body_site in ['NA', 'unknown', '', 'no_data', 'None']:
+        elif body_site in ['NA', 'unknown', '', 'no_data', 'None', 'Unknown']:
```
mortonjt: Personally, I would prefer `body_site.lower() in ['na', 'unknown', '', 'no_data', 'none']`

But this really doesn't matter.  I think it is good to merge.

squirrelo: Actually, I agree. Would make maintenance much easier, and I don't think case matters here, correct?

EmbrietteH: I think case does matter. AG round 1-22 processing failed in notebook 4 (metadata); when I added Unknown, uppercase, to my local file, the notebook ran successfully.

squirrelo: What we mean is does `Unknown` mean something different to `unknown` which means something different than `UNKNOWN`. If this is the case, this pull request is fine. If they all mean the same thing, converting to lowercase and then doing the comparison means we don't have to worry about what case the data actually is.

EmbrietteH: Since I've actually tested this specific fix, I'd rather merge this now. Let's revisit this later and we can change it to Jamie's suggestion if we think that is the way to go. Thanks!

wasade: Agree

PR #182, opened 2015-12-10
Introduces the single bodysite notebook

JWDebelius:
Creates a notebook for generating the single sample per bodysite packaged results.

PR #185, opened 2015-12-30
Per sample alpha plots

JWDebelius:
~~Depends on #182 for `add_alpha_diversity`~~

Generates the per sample alpha diversity plots:
shannon_1k_example


wasade:
Executing as a function in 09 would save a lot of overhead

JWDebelius:
Okay, I'll go back to the function version. I think it's easier to test that way.

JWDebelius:
After much work, I think I've gotten this to pass!

I had to drop the latex import, so its rendering in Ariel (example is Helvetica, but I don't think anyone will notice).
Also, I ran into the import issue. With the new (I think) version of matplotlib, it's necessary to specify the backend. I may have been a bit over zealous with that.

@wasade, @squirrelo, @mortonjt, @EmbrietteH: Could you please review?

americangut/per_sample.py
```diff
@@ -515,3 +577,132 @@ def stage_per_sample_specific_statics(opts, sample_ids):
...
+    stating this fact.
+
+    If the sample is present, a matplotlib figure with the alpha diversity
+    distribution and a line indicating the sample value is returned.
+
+    If debug is passed, the following parameters are returned:
+        group : str
+            The value of the `group_field` for the sample
+        group_alpha : ndarray
+            The alpha diversity values assoicated with the group
```
mortonjt: assoicated -> associated

americangut/per_sample.py
```diff
@@ -515,3 +577,132 @@ def stage_per_sample_specific_statics(opts, sample_ids):
...
+
+def _plot_alpha(sample, alpha_map, alpha_field, group_field='SIMPLE_BODY_SITE',
+                output_dir=None, xlabel=None, fp=None, debug=False):
+    """Generates a distrbution plot for the data
+
+    Parameters
+    ----------
+    sample : str
+        The sample ID to be plotted
+    alpha_map_fp : str, dataframe
```
mortonjt: preferable to have dataframe -> pd.DataFrame

americangut/per_sample.py
```diff
@@ -247,6 +258,57 @@ def taxa_summaries(opts, sample_ids):
...
+        A dict of relevant opts.
+    sample_ids : Iterable of str
+        A list of sample IDs of interest
+
+    Returns
+    -------
+    dict
+        A dict containing each sample ID and any errors observed or None if
+        no error was observed for the sample. {str: str or None}
+    """
```
mortonjt: Consider rewording? Keys are sampleIDs, values are error messages?

americangut/per_sample.py
```diff
@@ -515,3 +577,132 @@ def stage_per_sample_specific_statics(opts, sample_ids):
...
+        diversity values.
+    group_field : str
+        Default is 'SIMPLE_BODY_SITE'. The name of the column in `alpha_map`
+        which provides the grouping for generating distribution plots.
+    output_dir : str
+        The location where the alpha diversity figures should be saved.
+    xlabel : str
+        Text describing the quantity on the x-axis.
+
+    Returns
```
wasade: why return if the return isn't used?

americangut/per_sample.py
```diff
@@ -515,3 +577,132 @@ def stage_per_sample_specific_statics(opts, sample_ids):
...
+    -------
```
mortonjt: This returns is a bit convoluted.

What do you think about returning the exact same return types

fig, group, group_alpha, sample_alpha, xlabel, error_message

And just setting the attributes to None if debug is specified and what not.

americangut/per_sample.py
```diff
@@ -515,3 +577,132 @@ def stage_per_sample_specific_statics(opts, sample_ids):
...
+        group : str
+            The value of the `group_field` for the sample
+        group_alpha : ndarray
+            The alpha diversity values assoicated with the group
+        sample_alpha : float
+            The alpha diversity for the sample
+        xlabel : str
+            The label used for the x-axis of the plot.
+
+    Raises
```
wasade: since this has known conditions where it raises, can you these be caught in the driving block above? an exception can then be noted in the `result` object in `alpha_plot`. not catching the exception will result in all of the samples in the block being evaluated being impacted too, so this would help isolate issues.

ipynb/primary-processing/08-collapse_samples.md
```diff
@@ -17,23 +19,46 @@ Let's make sure we have the paths we need.
...
 
-First, we're going to operate on rarefied data again.
+We're going to start by generating per-bodysite mapping files with alpha diversity attached.
 
 ```python
 >>> low_depth, high_depth = agenv.get_rarefaction_depth()
 ```
 
 ```python
+>>> for trim in ['pgp-hmp-gg-100nt', 'notrim']:
```
wasade: should that also have "ag" in it?

wasade:
few small quick comments, :+1: otherwise

americangut/per_sample.py
JWDebelius: This is the documentation is modeled after every other function in this library.

americangut/per_sample.py
JWDebelius: Inline debugging

americangut/per_sample.py
JWDebelius: Because having multiple levels of returns means not having a tuple of Nones. This simplifies the input/output and targets whats needed.

JWDebelius:
Comments mostly addressed.

americangut/per_sample.py
```diff
@@ -515,3 +586,112 @@ def stage_per_sample_specific_statics(opts, sample_ids):
...
+    -------
+    If the sample is present, a matplotlib figure with the alpha diversity
+    distribution and a line indicating the sample value is returned. If a
+    file path is specified, the figure will be saved at the filepath instead
+    of returning.
+
+    If debug is passed, the following parameters are returned:
+        group : str
+            The value of the `group_field` for the sample
+        group_alpha : ndarray
```
mortonjt: ndarray -> np.ndarray ?

mortonjt:
One more minor comment.  Otherwise :+1:

americangut/per_sample.py
JWDebelius: Ive not seen that notation in doc strings before, typically. Could you direct me to the standard you're using?

Thanks

mortonjt: See bullet 9 [here](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections)

It's good practice to add the module name in front of imported functions/objects

mortonjt: Although the documentation guidelines don't really make this clear, so I think it is safe to ignore.

josenavas: just as a note, we started using @mortonjt 's suggestion in Qiita. I think we should start using such notation in any new code as it helps new developers unfamiliar with the code to get up to speed.

Issue #186, opened 2016-01-06
Analysis Summary Pipeline


jwdebelius:
I'd like to create something similar to the primary processing block, #161, for analysis, with the idea that the backend could be transferred over to Bokeh or similar later.

I think the easiest way to accomplish this might be a build a data dictionary backend, that would let people operate on the metadata, and then a holding object for the classes able to interact with that object. Test could go on top.

This way, we can have light weight, individual notebooks for each analysis step, and hopefully just switch  out plotting code at some point.

I see the steps as:
* Data Dictionary Objects
     * [ ] Parent Question Object (#188)
     * [ ] Categorical Question Objects (Categorical, Clinical, Frequency); (#192)
     * [ ] Boolean Question Objects (Bool, Multiple response); (#193)
     * [ ] Continous Questions (#194)
* [ ] Data handling Object (#195)
* [ ] Data Dictionary for Ag Questions
* [ ] Univariate Alpha diversity notebook with effect size
* [ ] Alpha diversity notebook for Scott Kelley's analysis
* [ ] Univariate Beta diversity notebook with effect size
* [ ] Univariate OTU notebook (different abundance and differential frequency)
* [ ] Univariate PICRUSt Notebook (have discussed this with @mortonjt)
* [ ] Multivariate alpha diversity
* [ ] Multivariate beta diversity

This may be redundant with other repositories, although I think the first step is unique to American Gut.

PR #188, opened 2016-01-21
Creates the question object

jwdebelius:
First of the data dictionary objects referenced in #186. 

The idea of the question object is to create a modular plug in system to facilitate analysis by other groups. Questions are prepopulated with features of the data including data type, any remapping functions which are used, and handy functions associated with common data types. The collection of objects can then be combined into a dictionary, and the dictionary be called as an underlying API for a set of easy, pretty functions.

This is a base class object for the question object.


JWDebelius:
Bumping this, because Im hoping to get a bunch of stuff in and this is necessary for all the question objects...

americangut/question/__init__.py
```diff
@@ -0,0 +1,21 @@
+import os
+
```
wasade: this logic is already in `americangut/__init__.py`

americangut/question/ag_question.py
```diff
@@ -0,0 +1,117 @@
+import inspect
+
+import numpy as np
+
+
+class AgQuestion:
+    """A base object class for handling American Gut Data dictionary entries"""
+    true_values = {'yes', 'y', 'true', 1, 1.0, 'Y', 'YES', 'Yes', True}
```
wasade: we're these already standardized?

americangut/question/__init__.py
JWDebelius: Should the __init__ document just be empty then? If so, sorry.

americangut/question/ag_question.py
```diff
@@ -0,0 +1,117 @@
...
+            The datatype in which the responses should be represented. (i.e.
+            float, int, str).
+        clean_name : str, optional
+            A nicer version of the way the column should be named.
+        remap : function
+            A function used to remap the data in the question.
+        free_response: bool, optional
+            Whether the question is a free response question or controlled
+            vocabulary
+        mimmarks : bool, optional
```
wasade: extra m?

this parameter and a few others are not described on the function definition

americangut/question/ag_question.py
```diff
@@ -0,0 +1,117 @@
...
+
+        self.type = 'Question'
+        if clean_name is None:
+            self.clean_name = name.replace('_', ' ').title()
+        else:
+            self.clean_name = clean_name
+
+        # Sets up
+        self.free_response = kwargs.get('free_response', False)
+        self.mimmarks = kwargs.get('mimmarks', False)
```
wasade: same, extra m

americangut/question/ag_question.py
wasade: i see below about the inquiry into kwargs, why not just setup defaults in the definition?

americangut/question/ag_question.py
```diff
@@ -0,0 +1,117 @@
...
+            raise ValueError('%s is not a column in the supplied map!'
+                             % self.name)
+
+    def remap_data_type(self, map_, watch=True):
+        """Makes sure the target column in map_ has the correct datatype
+
+        map_ : DataFrame
+            A pandas dataframe containing the column described by the question
+            name.
+        watch : bool
```
wasade: would it be possible to get a specific usecase for this? i'm not sure i understand what its use is

americangut/question/ag_question.py
```diff
@@ -0,0 +1,117 @@
...
+        else:
+            def remap_(x):
+                return self.dtype(x)
+
+        map_[self.name] = map_[self.name].apply(remap_).astype(self.dtype)
+        map_.replace('nan', np.nan, inplace=True)
+
+        if watch and self.type in {'Categorical', 'Multiple', 'Clinical',
+                                   'Bool', 'Frequency'}:
+            self._update_order(remap_)
```
wasade: this function does not exist

americangut/question/__init__.py
wasade: ya, that should be fine

americangut/question/ag_question.py
JWDebelius: I can trim, but there are multiple paths that might end up being cast to yes/no (Like, if you have a Yes/No/Not sure and then you drop the `Not sure` and want to treat it as bool.

americangut/question/ag_question.py
JWDebelius: The functionality gets leveraged in the Categorical subclass.
And example comes up in the function, [remove ambiguity](https://github.com/biocore/American-Gut/pull/192/files#diff-062a13454b23b479686ccaff0747068eR80). Here, you want to cast the data to the correct datatype, and then remove ambiguous entries. (In case, maybe, ambiguous entries are coded as -1, or something). I may also be over casting to the data type.

americangut/question/ag_question.py
JWDebelius: The logic is convuluted, but its a function in categorical questions and children. Other option is to split this out, and write one version of this function for categorical questions and another for continuous and children.

wasade: no, i mean the function does not appear to be defined anywhere

JWDebelius: https://github.com/biocore/American-Gut/pull/192/files#diff-062a13454b23b479686ccaff0747068eR69

americangut/question/ag_question.py
wasade: no, i mean didn't we aggressively standardize the metadata back in oct/nov? or was that standardization incomplete?

JWDebelius: Standardization was complete. The problem comes from trying to remap values here. So, if I remap the values as `'yes'` and `'no'` for one thing, and then decide later that I want to cast the column to bool values, I'd either have to reload the data (which might mean losing other work), or know that `'yes'` and `'no'` are the same as `'true'` and `'false'` for a `bool` column.

americangut/question/ag_question.py
wasade: Merging this PR means that there will be code that doesn't work in master. Why isn't the logic here just part of #192?

There is a problem with the object hierarchy. If this base class should not be instantiated, then it should be described as such or possibly be an ABC. If it is valid to instantiate the object, then it should implement `_update_order` at least to the level of throwing `NotImplementedError` as that is more informative than a `NameError`.

americangut/question/ag_question.py
```diff
@@ -0,0 +1,117 @@
...
+                    except:
+                        return np.nan
+        else:
+            def remap_(x):
+                return self.dtype(x)
+
+        map_[self.name] = map_[self.name].apply(remap_).astype(self.dtype)
+        map_.replace('nan', np.nan, inplace=True)
+
+        if watch and self.type in {'Categorical', 'Multiple', 'Clinical',
```
wasade: Does this indicate that the base class has strong knowledge of derived classes?

americangut/question/ag_question.py
wasade: I'm still totally lost as to what this is or does...

JWDebelius: For categorical variables, the question object monitors the groups present in the data as an order object. Changes in the responses are tracked using an `earlier_order` property, which describes past states of the `order` in a hope to give some kind of traceback of how the data has been modified. But, as I was modifying the order, the casting to dtype as part of some other function didn't seem useful to track, as each entry was supposed to correspond to one function call.

I may want to entirely rethink the logging system, though, and whether or not I need one...

I suppose I could impose an order with an error, where the data must be the correct type to be passed through the function, which implies that datatype has already been cast...

americangut/question/ag_question.py
JWDebelius: The base class can be instantiated. It's just really limited, and probably won't appear that way in the actual data dictionary function where these are being used.

I'm struggling with this, because there aren't great, clean ways to handle the logic. 

1. Break the method apart, and treat continuous and categorical data differently. This solves the inheritance problem, but code gets replicated.
2. Move the `_update_order` function into `AgQuestion`. Give it the logic to `update_order` depending on the type. This means updating the parent object every time you need to create a new derivative type, if you want it to behave correctly. That doesn't seem like the right solution.

wasade: Let me get this straight:

- an `AgQuestion` has a type, i.e., categorical
- an `AgQuestion` has a dtype, i.e., float
- an `AgQuestion` corresponds to an entry in the data dictionary

Shouldn't it be immutable since the data dictionary is immutable? Why would
a categories dtype change? Is there functionality that a derived class has
that doesn't make sense for all derived classes? What does "updating the
parent object" mean -- is this a singleton pattern? Are derivative types
created dynamically, if so, under what circumstances since the data
dictionary is static?

I'm sorry if I'm just totally lost here. This feels overly complex, but I
also don't fully understand what problem it is solving so it's hard to
evaluate. Can another developer review please?

JWDebelius: I'm really confused, as well. Because the issue here has less to do with the `dtype` parameter, and more to do with a difference between the two inherieted classes and the fact that I'm not sure you can modify a function in an inherited class and keep the same API.

There are two main types of questions: `AgCategorical` and `AgContinuous`. These are both derived from `AgQuestion`. Categorical questions track the groups in the data, using the `order` and `extremes` parameter. These are dynamic parameters, which are updated in conjunction with the data in the map. The `_update_order` parameter does nothing to the `AgQuestion.dtype`. It keeps the `order` and `extremes` up to date with the data in the map.

americangut/question/ag_question.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def check_map(self, map_):
+        """Checks the group exists in the metadata
+
+        Parameters
+        ----------
+        map_ : DataFrame
+            A pandas object containing the data to be analyzed. The
+            Question `name` should be a column in the `map_`.
+
```
mortonjt: Would it help to have a raises section here?

americangut/question/ag_question.py
```diff
@@ -0,0 +1,114 @@
...
+            raise ValueError('%s is not a column in the supplied map!'
+                             % self.name)
+
+    def remap_data_type(self, map_):
+        """Makes sure the target column in map_ has the correct datatype
+
+        map_ : DataFrame
+            A pandas dataframe containing the column described by the question
+            name.
+
```
mortonjt: Likewise here

americangut/question/ag_question.py
```diff
@@ -0,0 +1,114 @@
...
+        self.mimarks = mimarks
+        self.ontology = ontology
+        self.remap_ = remap
+
+    def check_map(self, map_):
+        """Checks the group exists in the metadata
+
+        Parameters
+        ----------
+        map_ : DataFrame
```
mortonjt: `DataFrame` or `pd.DataFrame`

In a previous PR, `pd.DataFrame` was used.

americangut/question/ag_question.py
```diff
@@ -0,0 +1,114 @@
...
+        remap : function
+            A function used to remap the data in the question.
+        free_response: bool, optional
+            Whether the question is a free response question or controlled
+            vocabulary
+        mimarks : bool, optional
+            If the question was a mimarks standard field
+        ontology : str, optional
+            The type of ontology, if any, which was used in the field value.
+
```
mortonjt: Same here about raises

tests/test_ag_question.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def test_init_class_error(self):
+        with self.assertRaises(TypeError):
+            AgQuestion(self.name, self.description, 'bool')
+
+    def test_init_clean_name_error(self):
+        with self.assertRaises(TypeError):
+            AgQuestion(self.name, self.description, self.dtype, ['Cats'])
+
+    def test_remap_data_type(self):
```
mortonjt: Want to have a test for the raises?

americangut/question/ag_question.py
```diff
@@ -0,0 +1,114 @@
...
+        else:
+            self.clean_name = clean_name
+
+        # Sets up
+        self.free_response = free_response
+        self.mimarks = mimarks
+        self.ontology = ontology
+        self.remap_ = remap
+
+    def check_map(self, map_):
```
mortonjt: Want to add a test case for this?

americangut/question/ag_question.py
JWDebelius: `DataFrame`

PR #189, opened 2016-01-22
Creates a notebook summarizing data

jwdebelius:
Basically what it says. PR for AG Table 1 in the paper.

### DIFF
--- a/ipynb/diversity-analysis/1-Summary-Table.md
+++ b/ipynb/diversity-analysis/1-Summary-Table.md
@@ -1,7 +1,7 @@
 
 The goal of this notebook is to provide demographic summaries for participants in the American Gut and associated projects. We look at metadata, and summarize the available information.
 
-The information generated here will be used for tables 1 of the American Gut paper.
+The information generated here will be used for table 1 of the American Gut paper.
 
 We'll start by importing the necessary libraries.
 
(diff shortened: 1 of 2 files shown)


ipynb/diversity-analysis/1-Summary-Table.md
```diff
@@ -0,0 +1,167 @@
+
+The goal of this notebook is to provide demographic summaries for participants in the American Gut and associated projects. We look at metadata, and summarize the available information.
+
+The information generated here will be used for tables 1 of the American Gut paper.
```
wasade: tables -> table

ipynb/diversity-analysis/1-Summary-Table.md
```diff
@@ -0,0 +1,167 @@
...
+import numpy as np
+import pandas as pd
+```
+
+Next, we're going to locate the mapping file and raw OTU table.
+
+
+```python
+processing_dir = os.path.abspath('../primary-processing/agp_processing/')
+map_fp = os.path.join(processing_dir, '01-raw/metadata.txt')
```
wasade: open question: should these paths be stored in `agenv` as well?

ipynb/diversity-analysis/1-Summary-Table.md
```diff
@@ -0,0 +1,167 @@
...
+md = pd.read_csv(map_fp, sep='\t', dtype=str, na_values=['NA', 'no_data', 'unknown', 'Unspecified', 'Unknown'])
+md.set_index('#SampleID', inplace=True)
+```
+
+We're also going to calculate the number of sequences per sample, and add that to our mapping file.
+
+
+```python
+otu_summary = !biom summarize-table -i $otu_fp
+seq_depth = pd.DataFrame(np.array([l.split(': ') for l in otu_summary[15:]]), columns=['#SampleID', 'counts'])
```
wasade: What about using the [method](https://github.com/biocore/biom-format/blob/master/biom/util.py#L311) directly? It would require loading the table, but the has the advantage of being invariant to the format of the output of `biom summarize-table`

ipynb/diversity-analysis/1-Summary-Table.md
```diff
@@ -0,0 +1,167 @@
...
+
+```python
+ag_participants.loc['Blank', 'AGP Participants'] = np.nan
+ag_participants.join(hmp_counts)
+```
+
+
+
+
+<div>
```
wasade: would it be possible to clear the output?

ipynb/diversity-analysis/1-Summary-Table.md
JWDebelius: It's probably cleaner. I'm not sure if its ofscicating.

Maybe it should be set up with a data directory and a results directory were it's assumed that the data directory is within the respository, and runs based not he absolute path of the repository (erroring if the data isn't there) and then a results directory which can be set based on the users needs.

If we do that, should agp_processing in primary-processing be added to a git ignore?

ipynb/diversity-analysis/1-Summary-Table.md
JWDebelius: I realize looking back that this isn't used here, and was in place for the demographics notebook #190. Will remove here, and update the other PR so it calls the method directly.

jwdebelius:
Closing in favor of re-analysis with Python 3.

PR #202, opened 2016-03-09
Updates so assert cells raise errors

JWDebelius:
An attempt to improve the way errors get handled, and prevent things
exiting with a 0.

Basically, switches the assert statements to raise Runtime errors instead of asserts.


josenavas:
@JWDebelius out of curiosity - how is this exactly going to help? As far as I know, asserts are still errors that are raised, so AssertionError and RuntimeError shouldn't behave differently, unless I'm missing something:
```
In [4]: assert 0 > 1, "Some error"
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-4-9b2ebc8c743d> in <module>()
----> 1 assert 0 > 1, "Some error"

AssertionError: Some error

In [5]: raise RuntimeError("Some error")
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-5-b2bc7e874145> in <module>()
----> 1 raise RuntimeError("Some error")

RuntimeError: Some error
```

Also, please remove the *.ipynb for ipython number 03 and re-add back the *.md file (I think you removed the wrong file).

wasade:
`AssertionError` is a subclass of `Exception`, so I agree, I'm not sure how
this changes anything.

What I suspect is going on is that some of the scripts are failing without
an exception. I do not think the issue is within the notebooks themselves

JWDebelius:
This was per a discussion with @EmbrietteH this afternoon. The problem is that scripts fail with an exit status of 0. Which is inconvenient.

squirrelo:
Does runipy run in optimized mode? It could be that the asserts are removed because of this, hence the failing silently. This should be more universally acceptable, so it fails even when run in python optimized mode.

PR #205, opened 2016-03-11
Converts alpha diversity to serial

JWDebelius:
Per discussion with @EmbrietteH, the alpha diversity call is converted to a serial, rather than parallel alpha diversity. This removes the issue with parallel alpha diversity potentially failing silently.

### DIFF
--- a/ipynb/primary-processing/05-alpha_diversity.md
+++ b/ipynb/primary-processing/05-alpha_diversity.md
@@ -57,20 +57,16 @@ Once we have the rarefactions, we can then compute the diversity of every sample
 
 ```python
 >>> for table, depth, rarefaction, multi_directory, adiv_directory, keybase in parameter_iterator():
-...     !parallel_alpha_diversity.py -i $multi_directory \
-...                                  -o $adiv_directory \
-...                                  -t $greengenes_tree \
-...                                  -m PD_whole_tree,chao1,observed_otus,shannon \
-...                                  -O $cpu_count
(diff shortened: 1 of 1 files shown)


josenavas:
:+1:

ipynb/primary-processing/05-alpha_diversity.md
```diff
@@ -57,20 +57,16 @@ Once we have the rarefactions, we can then compute the diversity of every sample
...
+...                         -o $adiv_directory \
+...                         -t $greengenes_tree \
+...                         -m PD_whole_tree,chao1,observed_otus,shannon
 ```
 
 We're going to aggregate the diversity calculations from the multiple rarefactions.
 
 ```python
 >>> for table, depth, rarefaction, multi_directory, adiv_directory, keybase in parameter_iterator():
-...     # sometimes QIIME leaves this directory, unsure why.
```
josenavas: For future reference - the times that I've seen QIIME leaving this directory is when the parallel call fails silently - so this may be suggesting that parallel alpha diversity was already failing sometimes and this piece of code was preventing detecting it in some other way.

wasade: i've also seen the directory left when it doesn't silently fail. it's
possible that the filesystems lag too much on barnacle

josenavas: Good point - forgot about NSF locks delay in barnacle

PR #207, opened 2016-04-18
Adding EBI pulldown script

josenavas:
This adds the script to pulldown the information from EBI.

@wasade I've added comments to the original bash script so it is easier to follow. Do you mind taking a look to make sure that they're accurate? Thanks!

### DIFF
--- a/scripts/get_new_samples_EBI.sh
+++ b/scripts/get_new_samples_EBI.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Grab any sample accessions that do not exist
+
+study_accession=ERP012803
+count=-1
+
+# Grab all the secondary sample accession numbers and file links from EBI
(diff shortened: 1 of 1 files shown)


wasade:
any idea what's up with the failure?

wasade:
Seems fine but I think it would be worth fixing travis?

josenavas:
Thanks @wasade ! Not sure what is going on with travis, will check later as I think it is unrelated to my changes.

wasade:
btw https://github.com/pypa/setuptools/issues/523

josenavas:
Thanks for the link @wasade , although I'm a bit confused cause it looks like they fixed the issue and travis is using setuptools 20.7, which should be a release with the fix included.

squirrelo:
[Known issue](https://github.com/pypa/setuptools/issues/544) with 20.7.0, which is the current version in conda. Downgrading to 20.6.7 fixes the issue.

EmbrietteH:
@josenavas what is the status of this?

PR #208, opened 2016-04-20
Generate taxa by summary data

squirrelo:
This is the first of a few PRs to generate the needed data for interactivity on the american gut site.

This PR generates the taxa summaries for each of the summary metadata categories we have (bmi, age, etc) that are generated for the static plot.


squirrelo:
Getting the same odd error as #207 but the rest is passing.

squirrelo:
Ready for review

squirrelo:
Ping @EmbrietteH @wasade @JWDebelius for review.

EmbrietteH:
I think it's good now regenerating what's in the static plot; in the future we may want to revisit adding capability to choose more categories...

americangut/per_category.py
```diff
@@ -0,0 +1,36 @@
...
+import americangut.notebook_environment as agenv
+from .util import collapse_full, collapse_taxonomy, get_existing_path
+
+
+def cat_taxa_summaries():
+    """Creates taxa summary files for each available summary category per site
+    """
+    paths = copy(agenv.paths['collapsed']['notrim']['1k'])
+    out_dir = get_existing_path(
+        agenv.path['populated-templates']['result-taxa'])
```
ekopylova: should this be ```agenv.paths``` ?

americangut/per_category.py
```diff
@@ -0,0 +1,36 @@
...
+    del paths['ag-biom']
```
ekopylova: why are you deleting this path?

ekopylova: ah because it's a biom table including all sites ..?

americangut/per_category.py
```diff
@@ -0,0 +1,36 @@
+from os.path import join
+from copy import copy
+
+from biom import load_table
+
+import americangut.notebook_environment as agenv
+from .util import collapse_full, collapse_taxonomy, get_existing_path
+
+
+def cat_taxa_summaries():
```
ekopylova: is there a test for this function?

americangut/per_category.py
```diff
@@ -0,0 +1,36 @@
...
+        table = load_table(get_existing_path(path))
+        if len(name.split('-')) == 2:
+            # Have entire cohort of samples for site, so need to get averages
+            table = collapse_full(table)
+        table = collapse_taxonomy(table)
+        ids = table.ids(axis='observation')
+        for col in table.ids():
+            if col == 'Unknown':
+                continue
+            cleaned_col = col.split('(')[0].strip().replace(' ', '_')
```
ekopylova: Is the text in braces not useful anymore? ex. Occasionally (1-2 times/week) -> Occasionally

americangut/per_category.py
```diff
@@ -0,0 +1,36 @@
...
+    paths = copy(agenv.paths['collapsed']['notrim']['1k'])
+    out_dir = get_existing_path(
+        agenv.path['populated-templates']['result-taxa'])
+    del paths['ag-biom']
+    for name, path in paths.items():
+        # consistent naming as stool for all participant items
+        name = name.replace('-biom', '').replace('fecal', 'stool')
+        table = load_table(get_existing_path(path))
+        if len(name.split('-')) == 2:
+            # Have entire cohort of samples for site, so need to get averages
```
ekopylova: get the medians?

americangut/per_category.py
ekopylova: never mind, this is only done for file naming ..

americangut/util.py
```diff
@@ -635,3 +636,58 @@ def get_single_id_lists(map_, depths):
...
+    ---------
+    [1] http://biom-format.org/documentation/table_objects.html
+    """
+    def collapse_f(id_, md):
+        return '; '.join(md['taxonomy'][:level + 1])
+    collapsed = _bt.collapse(collapse_f, axis='observation', norm=False)
+    return collapsed
+
+
+def collapse_full(_bt):
```
ekopylova: is it not possible to pass a collapse function to ```table.collapse()``` here? ex. as done qiime's ```collapse_samples.py``` ```_collapse_to_median```

americangut/per_category.py
squirrelo: correct

americangut/per_category.py
squirrelo: yeah, makes the file names managable

americangut/util.py
squirrelo: I tried that, but they need to collapse by sample not by taxa, and I don't think that's possible.

americangut/per_category.py
squirrelo: yeah, fixed

americangut/per_category.py
squirrelo: fixed

americangut/util.py
ekopylova: so in ```collapse_samples.py``` they don't collapse by samples?

squirrelo: sorry, collapse by sample data, not by taxa metadata. I'll dig into it and see if it's possible.

americangut/per_category.py
squirrelo: added

squirrelo:
Comments addressed, ready to go.

americangut/util.py
wasade: You might be able to do it with `collapse` but expressing `median` in that context will be weird as it would need to be a rolling median -- iirc, you can control the function that describes how values are smashed together, or it's possible we took that out as there was performance overhead. But, I think there is a more direct route:

```python
return Table([np.median(v) for v in _bt.iter_data(axis='observation')], 
             _bt.ids(axis='observation'),
             observation_metadata=_bt.metadata(axis='observation'))
```

Unfortunately, because of the rolling issue, you can't do the "cleanest" thing which would be a `reduce`:

```python
# won't work... :(
return Table.reduce(np.median, axis='observation')
```

wasade: ...at one point, we had discussed adding in `Table.apply` but that didn't make it

wasade:
Couple of small comments, :+1: otherwise

squirrelo:
png @ekopylova

americangut/util.py
squirrelo: little more complicated than the given return, but greatly simplified. Thanks.

ekopylova:
Thanks, looks good!

PR #209, opened 2016-04-20
Generate alpha diversity graphs for each metadata category of interest

squirrelo:
This generates the highlighted category alpha diversity graphs, allowing comparison on the interactive page. 

Requires #208


squirrelo:
Ready for review

squirrelo:
One test passed one failed due to timeout. Should still be good to review.

squirrelo:
ping @EmbrietteH @ekopylova

squirrelo:
ping @EmbrietteH @ekopylova

squirrelo:
ping @EmbrietteH @ekopylova

squirrelo:
ping @EmbrietteH @ekopylova

americangut/per_category.py
```diff
@@ -26,11 +28,83 @@ def cat_taxa_summaries():
...
             filename = '-'.join([name, cleaned_col]) + '.txt'
 
             with open(join(out_dir, filename), 'w') as f:
                 for otu, val in zip(ids, table.data(col)):
                     if val == 0.0:
                         continue
                     f.write('%s\t%s\n' % (otu, str(val)))
+
+
+def cat_alpha_plots():
```
ekopylova: no unit test?

americangut/per_category.py
```diff
@@ -26,11 +28,83 @@ def cat_taxa_summaries():
...
+                sample_name = group.split('(')[0].strip().replace(
+                    ' ', '_').rstrip('+')
+                pd_samps = alpha_map.loc[cat_groups[group], 'PD_whole_tree_1k']
+                pd_mean = pd_samps.mean()
+                pd_stdev = pd_samps.std()
+                shannon_samps = alpha_map.loc[cat_groups[group], 'shannon_1k']
+                sh_mean = shannon_samps.mean()
+                sh_stdev = shannon_samps.std()
+
+                # Add the new group row row to the metadata
```
ekopylova: row

americangut/results_utils.py
```diff
@@ -7,6 +7,8 @@
 from collections import defaultdict
 from functools import partial
 
+import matplotlib.pyplot as plt
```
ekopylova: This works fine on a local system but can fail on remote if forwarding is not available. Does it make sense to call ```plt.switch_backend('egg')``` after this line (or chose the backend before importing pyplot) to avoid that error?

americangut/results_utils.py
```diff
@@ -595,3 +597,144 @@ def write_bloom_fasta(unique_counts, output_file, abundance_threshold):
...
+            x, y = ax.lines[0].get_xydata().T
+            m = (highlight_range[0] < x) & (x < highlight_range[1])
+            ax.fill_between(x[m], 0, y[m], alpha=.15, color=group_color)
+            # Fill between changes axis ever so slightly, so need to reset
+            # back to what we want.
+            ax.set_xticks(xticks)
+
+        # Adds text describing the sample or category
+        if categorical:
+            text = '\n\n%s:\t%1.1f' % (sample, sample_alpha)
```
ekopylova: The .png graphs that I generated contain the ```[]``` symbol after the colon. Perhaps the '\t' isn't being read properly?

ekopylova:
lualatex seems to be a dependency but don't see it mentioned anywhere

ekopylova:
@squirrelo a few comments

americangut/results_utils.py
squirrelo: now set for agg rendering.

squirrelo:
The lualatex is needed for rendering the barcode results PDFs.

americangut/results_utils.py
squirrelo: Issue with tabs not being read correctly. Fixed now.

americangut/per_category.py
squirrelo: This generates images, and the core function generating the image is tested, so I'm not sure how to unit test this. Do you have suggestions?

ekopylova: Hmm how about computing the md5 (ex. ```self.assertFileMD5()``` in [this](https://grass.osgeo.org/grass72/manuals/libpython/gunittest.html) package)? Otherwise simply verifying your output file (shannon_path, pd_path) exists as you expect?

squirrelo: Hesitant to do MD5 because it could be sliiiightly rendered differently depending on the system tests are run on. Checking file is output would work though, and also makes sure the proper naming for files is there so the website can use them. Adding that.

squirrelo:
Test added for file names, but there is not a way to actually test the graph itself.

squirrelo:
Travis is not behaving for testing the rendering, and given that there are no other tests for rendering in the codebase, how much of a priority/how imperative is it that we have the render tests, given we're not \ checking the figure that is rendered and instead just the filename.

josenavas:
Just checked the error really fast. It looks like the error is saying that the label "70+" is not in the axis. I'm unsure on how the figure is generated and I understand that systems can change the output figure a bit. However, I think that changing the axis is not a small change and it should be ensure that the axis are as we expect. It may be a bit more helpful to see an example of the image generated attached in here.

squirrelo:
That error is the latest in a long string of errors, mostly having to do with the latex not rendering correctly or at all. Only the latest non-working commit is in the PR for brevity's sake.

squirrelo:
Here is an example graph:
pd_stool-age-30s

squirrelo:
Aaaand now tests are passing. Ok, cool.

josenavas:
Code looks good - however, before merging, I'm gonna pulldown this code and test it using the latest AG round

squirrelo:
ping @josenavas mergable?

josenavas:
No it's not:
```
Exception: Traceback (most recent call last):
  File "/home/jona1883/software/American-Gut/americangut/parallel.py", line 27, in run_functor
    return functor(*args, **kwargs)
  File "/home/jona1883/software/American-Gut/americangut/per_sample.py", line 117, in sample_type_processor
    return merge_error_reports(*[f(opts, ids) for f in functions])
  File "/home/jona1883/software/American-Gut/americangut/per_sample.py", line 307, in alpha_plot
    fp=shannon_path)
  File "/home/jona1883/software/American-Gut/americangut/results_utils.py", line 730, in plot_alpha
    sample_alpha, group_alpha.mean()).expandtabs()
AttributeError: 'tuple' object has no attribute 'expandtabs'
```
I though you told me that you did not modify the plot alpha function?

squirrelo:
The only change is to prepend the alpha diversity type, which we discussed offline. Investigating now.

squirrelo:
Fixed the issue. Was missing parenthesis.

PR #215, opened 2016-09-12
Justin's notebooks

jwdebelius:
Deblur notebooks from @jgeier.


wasade:
should this go here given that it is specific to data processed on barnacle?

jwdebelius:
This was associated with the notebook that we expect to go in the supplement. Again, I'm still not totally clear if we need to put paper analyses in this repository, or if we need another structure.

jgeier:
You have the wrong person.

